PR #1, opened 2018-03-22
[FIX] runbot: Add missing HTML attributes aria-*

luisg123v:
Some HTML attributes were missing, and are required to ensure good
compatibility with screen readers.

The following attributes are included:
- `aria-expanded`: this should be present in all dropdown menus
- `aria-label`: this should be present in all links and buttons which
  don't show text, but icons or images.


kujiu:
Hello,

Thanks for your work !

Have a nice day.

PR #2, opened 2018-03-23
merge bot

xmo-odoo:
# Questions:

- how to specify message for merge commit?

  al opinion: as is (`{pr.title}\n\n{pr.message}`) is fine, it's up to the author and reviewer to write those properly; note this does not apply to single-commit PRs (unless they're -rebase)

# todo
- [x] rebase & merge instead of straight merge

     Still needs a regular-merge mode (-rebase?) for forward-port so we don't get multi-merge (merge new into old then have mergebot merge new-old into new). Maybe if last commit has two parents rebase (and ff) it directly, and if only one then merge instead? That way the baseline is to create a branch B.fp, PR(from=B.fp, into=A) and if that passes it'll just merge B into A, if it doesn't the porter can add a head commit merging A into B, fixing what needs to be fixed, and *that commit* will be ~squashed as a merge commit by the mergebot.
- [x] signed webhooks: `body.repository.full_name` should exist on all webhook payloads and provide the "source" repo name ~~, however how to handle webhook-only repos? (e.g. odoo-dev repos are not managed but should send `status` hooks to the bot)~~ (fix runbot to properly send statuses on PRs instead)

  ~~can't be done, jsonrequest doesn't save the raw request body so we can't hash it to check~~

  having merged odoo/odoo#24107 means it could be done assuming that gets deployed on runbot (and we're fine requiring it)
- [x] delegate reviewers can trigger `retry`
- [x] should priority=0 (urgent) cancel the current staging and bypass reviews
- [x] automatically insert `closes #x` in commit message
- [x] handle "approved" github reviews, treat them as r+
- [x] post PR state using labels
- [x] labels should not be added on PRs created merged/closed (through PR sync) to avoid API churn
- [x] check PRs in p=0 state=error are not selected
- [x] when a staged PR is closed, cancel the staging
- [x] remove autosquash override, decide on squashing when merging

note: running the test suite locally (against a mock github) takes ~3mn, running the test suite remotely (against actual github repos & API) takes ~5h.


runbot_merge/models/pull_requests.py
```diff
@@ -0,0 +1,875 @@
...
+                    logger.info("Created staging %s", st)
+
+        Repos = self.env['runbot_merge.repository']
+        ghs = {}
+        self.env.cr.execute("""
+        SELECT
+            t.repository as repo_id,
+            t.pull_request as pr_number,
+            array_agg(t.id) as ids,
+            array_agg(t.state_from) as state_from,
```
KangOl: As you are only interested with first (chronologically) element (and last for `state_to`), you should order states with
```sql
array_agg(t.state_from order by id) as state_from,
array_agg(t.state_to order by id) as state_to
```

You can even get the correct element
```sql
(array_agg(t.state_from order by id))[1] as state_from,
(array_agg(t.state_to order by id desc))[1] as state_to
```

runbot_merge/controllers/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+    if 'pull_request' not in event['issue']:
+        return "issue comment, ignoring"
+
+    _logger.info(
+        'comment: %s %s:%s "%s"',
+        event['sender']['login'],
+        event['repository']['full_name'],
+        event['issue']['number'],
+        event['comment']['body'],
+    )
```
xmo-odoo: Put that after partner-check to avoid redundant spam.

runbot_merge/controllers/__init__.py
```diff
@@ -0,0 +1,258 @@
...
+
+def handle_review(event):
+    env = request.env(user=1)
+
+    partner = env['res.partner'].search([('github_login', '=', event['review']['user']['login'])])
+    if not partner:
+        _logger.info('ignoring comment from %s: not in system', event['review']['user']['login'])
+        return 'ignored'
+
+    pr = env['runbot_merge.pull_requests'].search([
```
xmo-odoo: Add acceptance log (same as comment)

PR #4, opened 2018-03-27
[FIX] runbot: Fixes for cleaned instances

moylop260:
- [REF] runbot: Remove used variable before assigned
  - `ValueError: <class 'UnboundLocalError'>: "local variable 'fd' referenced before assignment" while evaluating 'model._cron()'`
- [REF] runbot: Consider empty returned values
  - if builds path is empty show the error `bad query:  ... WHERE dest IN ()`
- [FIX] runbot: Fix KeyError if there are sticky branches without builds

### DIFF
--- a/runbot/common.py
+++ b/runbot/common.py
@@ -46,7 +46,6 @@ def locked(filename):
     try:
         fd = os.open(filename, os.O_CREAT | os.O_RDWR, 0o600)
     except OSError:
-        os.close(fd)
         return False
     try:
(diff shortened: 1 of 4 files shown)


moylop260:
@mart-e I just have created this PR with 3 fixes of runbot for special cases.
Take them if are good for you.

d-fence:
Merged here:
dcdd16038b88ec2fee551ba756dcbb63c924ba41
c9de61a320828ca21cef2885c6858f49ddfbb031
and 1e7edad85af80251bb6e0caae02831f1f669cf51

Thanks for the catching and fixing !

PR #5, opened 2018-03-27
[REF] runbot: hook - Find the right repo from payload

moylop260:
Work the task from TODO comment.

This change allow use a global webhook from github based on the organization without use a configuration for each repository.

I means, you should configure a global webook for all repository from the organization:
 - https://github.com/organizations/Vauxoo/settings/hooks

Instead of use a configuration for each project:
 - https://github.com/Vauxoo/project1/settings/hooks
 - https://github.com/Vauxoo/project2/settings/hooks
 - https://github.com/Vauxoo/project3/settings/hooks
 - https://github.com/Vauxoo/project4/settings/hooks
...

### DIFF
--- a/runbot/controllers/hook.py
+++ b/runbot/controllers/hook.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import datetime
+import json
 
 from odoo import http, tools
 from odoo.http import request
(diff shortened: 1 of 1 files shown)


moylop260:
@mart-e FYI this is a old pending `TODO`
For organizations with many projects is a lot helpful.

PR #6, opened 2018-03-27
[IMP] runbot: Add running logs button to show alive logs

moylop260:
This log is very important if you have a error in the frontend but that is not showed well.

cc @mart-e


Xavier-Do:
With the new config/step system, the running log link is added automatically.

PR #7, opened 2018-03-28
[FIX] runbot: py3 fixes

moylop260:
- Decode error
 - iteritems is not a attribute of dictionary anymore


d-fence:
Thanks again @moylop260 
Those py3 / v11 conversion issues flew under the radar.

PR #8, opened 2018-03-30
Master trigger enterprise build moc

d-fence:
Force a rebuild when a reverse dependency commit is detected.
e.g. when a new commit in community is fetched, force a rebuild of the latest commit in enterprise.

Use case: make it easier to detect a commit that makes the tests fail in another depending repository

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -415,7 +415,6 @@ class runbot_build(models.Model):
                 else:
                     v['job'] = jobs[jobs.index(build.job) + 1]
                 build.write(v)
-            build.refresh()
 
             # run job
             pid = None
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -59,6 +59,9 @@ class runbot_build(models.Model):
                                      ('fuzzy', 'Fuzzy - common ancestor found'),
                                      ('default', 'No match found - defaults to master')],
                                     string='Server branch matching')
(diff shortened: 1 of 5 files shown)


runbot/models/build.py
```diff
@@ -59,6 +59,9 @@ class runbot_build(models.Model):
                                      ('fuzzy', 'Fuzzy - common ancestor found'),
                                      ('default', 'No match found - defaults to master')],
                                     string='Server branch matching')
+    revdep_build_ids = fields.Many2many('runbot.build', 'runbot_rev_dep_builds',
```
antonylesuisse: Wha about dep_build_ids

antonylesuisse: Nope i get it rev == resverse

runbot/templates/build.xml
```diff
@@ -17,7 +17,27 @@
 
         <t t-if="bu['server_match'] in ('default', 'fuzzy')">
             <i class="text-warning fa fa-question-circle fa-fw"
-               title="Server branch cannot be determined exactly. Please use naming convention '9.0-my-branch' to build with '9.0' server branch."/>
+                title="Server branch cannot be determined exactly. Please use naming convention '9.0-my-branch' to build with '9.0' server branch."/>
+        </t>
+        <t t-if="bu['revdep_build_ids']">
+            <table class="table-condensed">
```
antonylesuisse: Display on same line as time

Issue #9, opened 2018-04-10
Builds are not killed when a branch or a repo is deleted


d-fence:
When a branch or a repo is deleted, their corresponding builds are not killed.

This causes two problems:
- The running builds are running indefinetly
- The _find_port method can return a port that is used by a "phantom" build still running. Leading to a traceback 'Address already in use'

PR #10, opened 2018-04-11
[FIX] runbot: kill builds when unlinking repo or branch

d-fence:
When a repo is deleted, the corresponding builds processes continues to
run indefinitely. Because of that, it may happens that the runbot launch
a new build using a port that is still in use by such a zombie build.

If this commit is applied, when unlinking a repo or a branch, the
corresponding builds are marked to be killed. Also, when searching for a
free port, the port is checked to avoid 'port already in use'
tracebacks.

It should fix https://github.com/odoo/runbot/issues/9

### DIFF
--- a/runbot/common.py
+++ b/runbot/common.py
@@ -21,6 +21,17 @@ def fqdn():
     return socket.getfqdn()
 
 
+def is_port_available(port):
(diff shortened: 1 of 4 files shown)


Xavier-Do:
Looks like it is outdated

PR #12, opened 2018-04-20
[IMP] runbot: choose the python version for coverage

d-fence:
When code coverage was processed the 'coverage' utility was called the
same way regardless of the Odoo version. That was the cause of two
problems:
1) In some OS packages, the 'coverage' executable was renamed to
'python-coverage' and 'python3-coverage'
2) Since version 11.0, Odoo needs python3

With this commit, the coverage module is called from python '-m'
argument and the python version is choosen from the Odoo executable
shebang.

### DIFF
--- a/runbot/common.py
+++ b/runbot/common.py
@@ -98,3 +98,14 @@ def local_pgadmin_cursor():
     finally:
         if cnx:
             cnx.close()
+
+def get_py_version(build):
+    """return the python name to use from build instance"""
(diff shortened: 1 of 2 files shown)


d-fence:
@rco-odoo as discussed, you can exclude code with a comment like this: "# pragma: no cover" as stated here: http://coverage.readthedocs.io/en/latest/excluding.html

PR #13, opened 2018-04-26
[REF] runbot: Skip build from commit message

moylop260:
This module skip a build based on the commit message.
Inspired by [Travis-CI#Skipping-a-build](https://docs.travis-ci.com/user/customizing-the-build#Skipping-a-build)

 - If you don’t want to run a build for a particular commit for any reason, add `[ci skip]` or `[skip ci]` to the git commit message.
 - Commits that have `[ci skip]` or `[skip ci]` anywhere in the commit messages are ignored by Travis CI.
 - Note that in case multiple commits are pushed together, the `[skip ci]` or `[ci skip]` takes effect only if present in the commit message of the HEAD commit.

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -15,6 +15,9 @@ from odoo.modules.module import get_module_resource
 from odoo.tools import config
 from ..common import fqdn, dt2time
 
+SKIP_WORDS = ['[ci skip]', '[skip ci]']
+SKIP_WORDS_RE = re.compile("|".join(map(re.escape, SKIP_WORDS)))
+
 _logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


d-fence:
Hello @moylop260 , is there any reason to create a build and skip it ?
Why not ignoring completely the commit in the runbot_repo._update_git method ?

moylop260:
Hello @d-fence
I delete the tmpdir from tearDown when the tests finish

d-fence:
Hello @moylop260 
Yes, you delete the tmpdir but the repo is cloned in the static dir of the runbot https://github.com/odoo/runbot/blob/9555370fa678a1cc7f8b2d9e99b8ba07c514b6ce/runbot/models/repo.py#L134 

So after the test, the cloned repo is still there. It's not that bad, because this dir is "gitignored " but it would be cleaner to remove that after the test.

moylop260:
Hello @d-fence 
You are right!

Fixed

moylop260:
I just merge your branch to this one and I have created a small change.
Thanks @d-fence

moylop260:
@d-fence If there is another change you just request me
Regards!

d-fence:
@moylop260 Hello. Yes, we discussed internally and we would like to change a little bit.
We would like those keywords:
[skip all] : same as [ci skip]
[skip tests] : a build object is created but the test jobs are skipped
[skip build] : a build object is created but the last job (build) is skipped

What do you think about that ?

moylop260:
I agree with you.

In order to have compatibility with others CI like as gitlab and travis... we should use `[ci skip]` for skip all since that they work this way.

`[skip tests]` and `[skip build]` is a good idea for runbot.
IMHO this should be other PR since that this PR is doing the initial work fine.

KangOl:
FYI, It break in case of mutliline commit message.

moylop260:
@kangOI Could you give us more details, please?

Do you have a traceback? (Could you attach it)
What is the commit message used?

For the tests we are using: 
 - `cimsg = "Testing body\n\n[ci skip]"`

But maybe we need to add another special case...

d-fence:
Finally removed from runbot.

PR #17, opened 2018-05-10
[FIX] runbot: Avoid infinite loop fetching

moylop260:
Using `repo.hook_time = None` and `repo.mode = hook` the cron run fetch command in an infinite-loop.

Using by default the current date the issue is fixed because the first command is a "git clone" if the path is not exists then the last time is now for first time.


mreficent:
@moylop260 I think this should be rebased

moylop260:
Rebased

PR #20, opened 2018-05-18
[IMP] runbot: kill older builds when a new one is created

d-fence:
When a new commit is found and a new build is created. When a user
pushes more commits in the same branch a few minutes later, it causes
more builds in testing phase, resulting in a CPU waste.

With this commit, previous builds in testing phase in non sticky
branches are killed when a new build is created.

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -205,6 +205,16 @@ class runbot_repo(models.Model):
                     builds_to_skip._skip(reason='New ref found')
                     if builds_to_skip:
                         build_info['sequence'] = builds_to_skip[0].sequence
+                    # testing builds are killed
+                    builds_to_kill = Build.search([
+                        ('branch_id', '=', branch.id),
+                        ('state', '=', 'testing'),
+                        ('committer', '=', committer)
+                    ])
+                    builds_to_kill.write({'state': 'deathrow'})
+                    for btk in builds_to_kill:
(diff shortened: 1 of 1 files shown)


runbot/models/repo.py
```diff
@@ -211,6 +211,16 @@ def _update_git(self):
                     builds_to_skip._skip(reason='New ref found')
                     if builds_to_skip:
                         build_info['sequence'] = builds_to_skip[0].sequence
+                    # testing builds are killed
+                    builds_to_kill = Build.search([
+                        ('branch_id', '=', branch.id),
+                        ('state', '=', 'testing'),
+                        ('commiter', '=', committer)
```
mreficent: Could we just kill it ONLY if the commit has been replaced?

I mean,

Case 1:
Add commit A to branch
While testing A, modify commit A by doing a rebase (thus replacing it by the modified commit) and push force
Expected: previous build deleted.

Case 2:
Add commit A to branch
While testing A, add new commit B to branch
Expected: no build is deleted.

d-fence: typo committer

KangOl:
I would only do this on non-sticky branches

d-fence:
That's the case https://github.com/odoo/runbot/blob/31209920c28afa544c9fa1516ac0c0c866089c14/runbot/models/repo.py#L200

PR #21, opened 2018-05-31
[IMP] runbot: update nginx template

d-fence:
When someone tries to log in an old runbot build that is not running
anymore, he lands on the runbot instance that was running the build.
Also, all the running builds are allowed on all runbot instances,
leading to the same behavior.

With this commit, only the builds that are running on the runbot
instance can be reached, others are defaulted to a 404.

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -289,10 +289,10 @@ class runbot_repo(models.Model):
         nginx_dir = os.path.join(self._root(), 'nginx')
         settings['nginx_dir'] = nginx_dir
         settings['re_escape'] = re.escape
+        settings['fqdn'] = fqdn()
         nginx_repos = self.search([('nginx', '=', True)], order='id')
         if nginx_repos:
-            builds = self.env['runbot.build'].search([('repo_id', 'in', nginx_repos.ids), ('state', '=', 'running')])
-            settings['builds'] = self.env['runbot.build'].browse(builds.ids)
+            settings['builds'] = self.env['runbot.build'].search([('repo_id', 'in', nginx_repos.ids), ('state', '=', 'running'), ('host', '=', fqdn())])
 
             nginx_config = self.env['ir.ui.view'].render_template("runbot.nginx_config", settings)
             os.makedirs(nginx_dir, exist_ok=True)
--- a/runbot/templates/nginx.xml
+++ b/runbot/templates/nginx.xml
@@ -52,6 +52,11 @@ server {
     location /longpolling { proxy_pass http://127.0.0.1:<t t-esc="build.port + 1"/>; }
 }
 </t>
+server {
+    listen 8080;
+    server_name ~.+\.<t t-raw="re_escape(fqdn)"/>$;
+    location / { return 404; }
+}
 }
       </template>
     </data>


d-fence:
@KangOl or @julienlegros can you have an eye on that ?

runbot/templates/nginx.xml
```diff
@@ -52,7 +52,12 @@ server {
     location /longpolling { proxy_pass http://127.0.0.1:<t t-esc="build.port + 1"/>; }
 }
 </t>
+server {
+    listen 8080;
+    server_name "~<t t-raw="re_escape(fqdn_re)"/>$";
```
KangOl: no need to be a regex. could simply be
```
  server_name *.<t t-raw="fqdn"/>;
```

runbot/templates/nginx.xml
```diff
@@ -52,7 +52,12 @@ server {
...
+    location /runbot { return 404; }
```
KangOl: location /

runbot/models/repo.py
```diff
@@ -289,9 +289,10 @@ def _reload_nginx(self):
         nginx_dir = os.path.join(self._root(), 'nginx')
         settings['nginx_dir'] = nginx_dir
         settings['re_escape'] = re.escape
+        settings['fqdn_re'] = '.*\.' + self.fqdn()
         nginx_repos = self.search([('nginx', '=', True)], order='id')
         if nginx_repos:
-            builds = self.env['runbot.build'].search([('repo_id', 'in', nginx_repos.ids), ('state', '=', 'running')])
+            builds = self.env['runbot.build'].search([('repo_id', 'in', nginx_repos.ids), ('state', '=', 'running'), ('host', '=', fqdn())])
             settings['builds'] = self.env['runbot.build'].browse(builds.ids)
```
KangOl: also, no need to rebrowse builds

runbot/templates/nginx.xml
```diff
@@ -52,7 +52,12 @@ server {
     location /longpolling { proxy_pass http://127.0.0.1:<t t-esc="build.port + 1"/>; }
 }
 </t>
+server {
+    listen 8080;
+    server_name "*.<t t-raw="fqdn"/>$";
```
KangOl: extra `$` at the end

KangOl:
Damn! My bad.

You actually MUST use a regex for the fallback `server_name`.
As build `server_name` are regex, if we define the fallback as wildcard, it will be processed first [0]

[0] https://nginx.org/en/docs/http/server_names.html

runbot/templates/nginx.xml
d-fence: Yes, it need to be a regex because glob patterns take precedence over regex in server blocks.
It means that this last server block will be matched before the server blocks for each specific builds running on the runbot instance and they will be unreachable.
I tested it.
as explained at the top of this page: https://nginx.org/en/docs/http/server_names.html

runbot/templates/nginx.xml
```diff
@@ -52,7 +52,12 @@ server {
     location /longpolling { proxy_pass http://127.0.0.1:<t t-esc="build.port + 1"/>; }
 }
 </t>
+server {
+    listen 8080;
+    server_name "~.+\.<t t-raw="fqdn"/>$";
```
KangOl: I'm not sure double-quoting it works

KangOl: And you should `re_escape(fqdn)`

d-fence: It runs on the test runbot but I will re_escape, sure

runbot/templates/nginx.xml
d-fence: @KangOl lol, We came to the same conclusion

runbot/templates/nginx.xml
d-fence: re_escape is needed, you are true, otherwise '.' will match anything. Sorry for the noise.

PR #24, opened 2018-06-07
[FIX] runbot: add recycle icon only once

d-fence:
Adding a unicode icon in the subject led to some issues. e.g. a user doesn't have any clue of what the icon really means. Also, the icon not the same everywhere, depending on the browser and the fonts installed.

With this commit, a field is added to differentiate build types. Using that information, an fa icon is displayed with the appropriate title description.

### DIFF
--- a/runbot/controllers/frontend.py
+++ b/runbot/controllers/frontend.py
@@ -35,6 +35,8 @@ class Runbot(http.Controller):
             'duplicate_of': build.duplicate_id if build.state == 'duplicate' else False,
             'coverage': build.coverage or build.branch_id.coverage,
             'revdep_build_ids': sorted(build.revdep_build_ids, key=lambda x: x.repo_id.name),
+            'build_type' : build.build_type,
+            'build_type_label': dict(build.fields_get('build_type', 'selection')['build_type']['selection']).get(build.build_type, build.build_type),
         }
 
     def _pending(self):
--- a/runbot/templates/dashboard.xml
+++ b/runbot/templates/dashboard.xml
@@ -61,6 +61,8 @@
                         </t>
                         <br/>
                         <i class="fa fa-envelope-o"></i>
+                        <t t-if="bu['build_type']=='scheduled'"><i class="fa fa-clock-o" t-att-title="bu['build_type_label']" t-att-aria-label="bu['build_type_label']"/></t>
+                        <t t-if="bu['build_type'] in ('rebuild', 'redirect')"><i class="fa fa-recycle" t-att-title="bu['build_type_label']" t-att-aria-label="bu['build_type_label']"/></t>
                         <a t-attf-href="https://#{repo['base']}/commit/#{bu['name']}"><t t-esc="bu['subject'][:32] + ('...' if bu['subject'][32:] else '') " t-att-title="bu['subject']"/></a>
                         <br/>
                         <t t-call="runbot.build_name"/> — <small><a t-attf-href="/runbot/build/{{bu['id']}}"><t t-esc="bu['dest']"/></a> on <t t-esc="bu['host']"/> <a t-if="bu['state'] == 'running'" t-attf-href="http://{{bu['domain']}}/?db={{bu['dest']}}-all"><i class="fa fa-sign-in"></i></a></small>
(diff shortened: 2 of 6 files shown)


KangOl:
While it's works, it should be nice to also strip other automatically prefixes like "⏰" added by recurrent builds.

An idea would be to have a method on build model
```python
def _tag(self, tag):
    know_tags = '⏰♻️'   # XXX get it from ir.config_parameter?
    self.subject = '%s %s' % (tag, self.subject.lstrip(know_tags))
```

d-fence:
@KangOl what do you think of that PR ?

runbot/templates/dashboard.xml
```diff
@@ -61,6 +61,8 @@
                         </t>
                         <br/>
                         <i class="fa fa-envelope-o"></i>
+                        <t t-if="bu['build_type']=='scheduled'"><i class="fa fa-clock" t-attf-title="bu['build_type']" t-attf-aria-label="bu['build_type']"/></t>
```
KangOl: title (and aria-label) will be `rebuild` instead of `This build is a rebuild`

runbot/controllers/frontend.py
```diff
@@ -35,6 +35,8 @@ def build_info(self, build):
             'duplicate_of': build.duplicate_id if build.state == 'duplicate' else False,
             'coverage': build.coverage or build.branch_id.coverage,
             'revdep_build_ids': sorted(build.revdep_build_ids, key=lambda x: x.repo_id.name),
+            'build_type' : build.build_type,
+            'build_type_label': [x[1] for x in build.fields_get('build_type','selection')['build_type']['selection'] if x[0] == build.build_type][0]
```
d-fence: @KangOl I didn't find better than that to get the selection label instead of the value.

KangOl: It's easier by converting selection to a dict
```python
dict(build.fields_get('build_type', 'selection')['build_type']['selection']).get(build.build_type, build.build_type)
```

d-fence: smart ! I change that Thanks

d-fence:
@antonylesuisse I know we already discussed this one but as I add a field, I prefer you see it before I merge.

KangOl:
I would also add `indirect` type [1] for builds generated by dependence update

[1] unless you have a better name...

d-fence:
@KangOl what about "autorebuild" for the builds generated by dependence update ?
Do you think about a particular fa icon ? I searched and nothing seems obvious.

KangOl:
https://fontawesome.com/icons/magic ?

d-fence:
I thought about it but it's not clear and too much Christmas stars

KangOl:
Then, choose the emoji you want...

runbot/templates/frontend.xml
```diff
@@ -149,6 +149,8 @@
                                   <t t-if="bu['state'] in ['running','done'] and bu['result'] in ['killed', 'manually_killed']"><t t-set="klass">killed</t></t>
                                   <td t-attf-class="{{klass}}">
                                      <t t-call="runbot.build_button"><t t-set="klass">btn-group-sm</t></t>
+                                     <t t-if="bu['build_type']=='scheduled'"><i class="fa fa-clock" t-attf-title="bu['build_type_label']" t-attf-aria-label="bu['build_type_label']"/></t>
+                                     <t t-if="bu['build_type'] in ('rebuild', 'indirect')"><i class="fa fa-recycle" t-attf-title="bu['build_type_label']" t-attf-aria-label="bu['build_type_label']"/></t>
```
d-fence: @KangOl at this time, I keep the same recycle icon, only the hover text changes

KangOl: That's works

KangOl: You may change the icon color...

mreficent:
https://fontawesome.com/icons/sync
https://fontawesome.com/icons/link
https://fontawesome.com/icons/play-circle
https://fontawesome.com/icons/project-diagram
https://fontawesome.com/icons/retweet
https://fontawesome.com/icons/code-branch

PR #27, opened 2018-06-11
[runbot] improve coverage handling

hbrunn:
[IMP] copy coverage flag when rebuilding
[IMP] omit during report build what we omitted during measuring


runbot/models/build.py
```diff
@@ -286,6 +286,7 @@ def _force(self, message=None):
                     'committer_email': build.committer_email,
                     'subject': u'♻️ %s' % build.subject,
                     'modules': build.modules,
+                    'coverage': build.coverage,
```
d-fence: Actually, it's not a feature that we want. The coverage builds takes a longer time than the other builds. That's the main reason of our choice to process coverage during the night. We don't want a reverse dependency build to trigger a coverage build.

hbrunn: will it be acceptable if I change this call to get the data from some `_force_rebuild(self, build, sequence)` function so that we can add this in a module cleanly?

runbot/models/build.py
```diff
@@ -771,20 +772,32 @@ def _job_20_test_all(self, build, lock_path, log_path):
...
+        dummy, mods = build._cmd()
+        available_modules = [
+            os.path.basename(os.path.dirname(a))
+            for a in (glob.glob(build._server('addons/*/__openerp__.py')) +
+                      glob.glob(build._server('addons/*/__manifest__.py')))
+        ]
+        bad_modules = set(available_modules) - set((mods or '').split(','))
+        return [
+            '--omit', ','.join(build._server('addons', m) for m in bad_modules)
+        ] if bad_modules else []
```
hbrunn: @d-fence I'm of a mind to add `_coverage_source()` and `_coverage_include()` here so that addons can fine tune what to measure. Will this be acceptable?

d-fence:
I'm currently working on the coverage improvement (see #25 ). It's really a work in progress, so I think that it would be better to wait that those little improvement are made. I appreciate your work.

runbot/models/build.py
d-fence: Unfortunately, it's not a feature that we need. As the runbot is more an internal tool that we share, we cannot include/maintain features that we will not use. I'm sorry.

hbrunn:
thanks for the info, I just come back from reviewing this ;-) I'll come back to this when #25 is merged

mreficent:
@hbrunn could you update your PR?

hbrunn:
I will, but from @d-fence I understand this code will stay a moving target for a while, so I prefer to wait until this is stable and then adapt my code

hbrunn:
conflicts fixed

daramousk:
From what I understand here you want to improve coverage by not including `__openerp__.py` and `__manifest__.py`, I was getting conflicts here so I rebased on top of master and it seems that they have implemented this functionality here https://github.com/odoo/runbot/commit/5761fd169445fc2cfdedcac69119c061b103c393#diff-5a9a0bdc452956a6a273746fda551875 (the `__openerp__.py` was not included)

At this point I am not sure if this will be merged and/or used.

hbrunn:
it won't, but if there are other people still using the pre docker runbot, they might want that

Xavier-Do:
Looks like this is outdated.

PR #37, opened 2018-09-11
[FIX] runbot_merge: avoid double closes message

mart-e:
If the message already contains the pr number, no need to add it again


xmo-odoo:
Outstanding!

PR #40, opened 2018-09-25
[IMP] runbot: add search and sort to frontend builds by branch

d-fence:
In the builds by branch frontend view:
* add the possibility to change the sort order of some columns
* add a search input that search in various build fields

### DIFF
--- a/runbot/controllers/frontend.py
+++ b/runbot/controllers/frontend.py
@@ -7,6 +7,7 @@ from odoo import http
 from odoo.addons.http_routing.models.ir_http import slug
 from odoo.addons.website.controllers.main import QueryURL
 from odoo.http import request
+from odoo.osv import expression
 from ..common import uniq_list, flatten, s2human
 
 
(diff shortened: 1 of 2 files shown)


runbot/templates/branch.xml
```diff
@@ -3,52 +3,79 @@
...
+                                    </div>
+                                    <div class="col-md-4">
+                                        <span class="pull-right"><t t-call="website.pager" /></span>
+                                    </div>
+                                </div>
+                            </div>
+                            <table class="table table-condensed table-stripped" style="table-layout: initial;">
+                                <thead>
+                                    <tr>
+                                        <th><a t-attf-href="?sorting=sequence&amp;sortdir={{ 'desc' if sortdir == '' else '' }}&amp;search={{ search if search else '' }}">Seq</a></th>
```
xmo-odoo: The sortdir bit is a bit odd: if my reading is correct, if currently sorting by date ascending clicking on sequence will make it sort by sequence descending?

Also the search bit probably needs to be HTML-escaped (I don't see it in the controller though I may have missed it), otherwise you can probably XSS people by sending them a link with a set search.

runbot/templates/branch.xml
```diff
@@ -3,52 +3,79 @@
...
-                    </table>
-                  </div>
-              </div>
+                    </div>
+                </t>
+                <t t-if="not builds">
+                    <div class="text-center">
+                        <h3>No builds found</h3>
+                    </div>
+                </t>
```
xmo-odoo: Maybe put that as a single row the width of the entire table instead? It might be less weird than seeing the entire table disappear.

runbot/controllers/frontend.py
```diff
@@ -333,17 +334,26 @@ def glances(self, refresh=None):
         return request.render("runbot.glances", qctx)
 
     @http.route(['/runbot/branch/<int:branch_id>', '/runbot/branch/<int:branch_id>/page/<int:page>'], website=True, auth='public', type='http')
-    def branch_builds(self, branch_id=None, search='', page=1, limit=50, refresh='', **kwargs):
+    def branch_builds(self, branch_id=None, search='', sorting='sequence', sortdir='desc', page=1, limit=50, refresh='', **kwargs):
         """ list builds of a runbot branch """
-        builds_count = request.env['runbot.build'].search_count([('branch_id','=',branch_id)])
+        limit = limit if limit <= 200 else 200
```
xmo-odoo: That was probably broken before, and will probably be broken after: I'm guessing there's no limit being provided anywhere at the moment, but if there is it's going to be a string, and so going to blow up. This only worked before (and likely only works now) because the default value is an integer.

PR #46, opened 2018-10-17
[FIX] runbot: build PR with corresponding community PR

d-fence:
When searching for a matching PR in a target repo, the match was made on
the 'base ref' which is the base branch on which the PR is based.

for example: PR 384155 should find a PR 27734 because their head ref are
the same. Instead of that, a match is made on their base branch.

This issue is annoying for the runbot_merge because it prevent some PR
of being merged.

With this commit, the match is made on the on the head ref.


runbot/models/build.py
```diff
@@ -133,8 +133,7 @@ def _get_closest_branch_name(self, target_repo_id):
 
         build = self
         branch, repo = build.branch_id, build.repo_id
-        pi = branch._get_pull_info()
-        name = pi['base']['ref'] if pi else branch.branch_name
+        name = branch.pull_head_name or branch.branch_name
```
mart-e: This is the name of the branch, e.g. "patch~1". Are we sure it won't match on another PR by somebody else using the same name? To check with the full label (e.g. mart-e:patch~1)

d-fence:
Thanks for the bug catch @xmo-odoo 
Could be fixed here: https://github.com/odoo/runbot/pull/47

tde-banana-odoo:
Merge ! We have staging branches ! Merge ! Merge !

d-fence:
@tde-banana-odoo No, the fix is bad as @mart-e emphased

tde-banana-odoo:
Maybe, but you are 100% yourself :heart:

runbot/models/branch.py
```diff
@@ -75,6 +76,13 @@ def _get_pull_head_name(self):
             if pi:
                 branch.pull_head_name = pi['head']['ref']
 
+    def _get_pull_head_label(self):
+        """lazy compute of the head label"""
+        for branch in self:
+            pi = self._get_pull_info()
+            if pi:
+                branch.pull_head_label = pi['head']['label']
+
```
d-fence: @xmo-odoo I know it's ugly but if I use a compute field, I think that it will lead to migration problems.
What do you think of that ?

xmo-odoo: Why not just alter pull_head_name to store the label instead of the head? Do we do anything useful with the pull_head_name at this point?

d-fence: That's true but existing PR are not going to be updated.

d-fence: I think that you are right @xmo-odoo . I finally pushed a cleaner/smaller code.

runbot/models/branch.py
```diff
@@ -72,8 +72,9 @@ def _get_pull_head_name(self):
         """compute pull head name"""
         for branch in self:
             pi = self._get_pull_info()
-            if pi:
-                branch.pull_head_name = pi['head']['ref']
+            if pi and not _re_patch.match(pi['head']['label']):
```
d-fence: @xmo-odoo I added a regex to avoid the 'patch-' issue

xmo-odoo: Looks good to me.

PR #51, opened 2018-10-30
[IMP] runbot: containerize builds

d-fence:
Use docker to containerize builds.

### DIFF
--- a/runbot/common.py
+++ b/runbot/common.py
@@ -34,28 +34,6 @@ def now():
     return time.strftime(DEFAULT_SERVER_DATETIME_FORMAT)
 
 
-def lock(filename):
-    fd = os.open(filename, os.O_CREAT | os.O_RDWR, 0o600)
-    if hasattr(os, 'set_inheritable'):
-        os.set_inheritable(fd, True)  # needed since pep-446
-    fcntl.lockf(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
-
-
-def locked(filename):
-    result = False
-    try:
-        fd = os.open(filename, os.O_CREAT | os.O_RDWR, 0o600)
-    except OSError:
-        return False
-    try:
-        fcntl.lockf(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
-    except OSError:  # since pep-3151 fcntl raises OSError and IOError is now an alias of OSError
-        result = True
-    finally:
-        os.close(fd)
-    return result
-
-
 def grep(filename, string):
     if os.path.isfile(filename):
         return open(filename).read().find(string) != -1
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -1,0 +1,175 @@
+# -*- coding: utf-8 -*-
+"""Containerize builds
+
+The docker image used for the build is always tagged like this:
+    odoo:runbot_tests
+This file contains helpers to containerize builds with Docker.
+When testing this file:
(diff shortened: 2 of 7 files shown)


d-fence:
@julienlegros I finally have a Dockerfile for all versions that seems to work.

antonylesuisse:
LGTM

runbot/common.py
```diff
@@ -14,6 +14,8 @@
 
 from odoo.tools.misc import DEFAULT_SERVER_DATETIME_FORMAT
 
+VERSION_RE = '^(version_info ?= ?\()(.+)(, ?)(.+)(, ?)(.+, ?){3}'
```
xmo-odoo: The parenthesisation seems excessive, most of the groups are unnecessary and the numbers should be matched using `\d+` rather than `.+`. Also should use rawstrings for safety. And could name the fields:

    r'^version_info ?= ?\((?P<major>\d+), ?(?P<minor>\d+), ?(.+, ?){3}'

runbot/common.py
```diff
@@ -99,13 +101,10 @@ def local_pgadmin_cursor():
...
-                if f.readline().strip().endswith('python3'):
-                    return 'python3'
-    return 'python'
-
+def get_odoo_version(release_path):
+    """Safely extract and return the odoo version from a release.py file"""
+    with open(release_path, 'r') as f:
+        d = f.read().replace('~','-')
+    m = re.search(VERSION_RE,d, re.MULTILINE)
+    version = m and '%s.%s' % (m.group(2).strip("'").strip('"'), m.group(4).strip('"').strip("'"))
```
xmo-odoo: Alternatively, we could use ast. I don't know that it would be an improvement but:
```python
mod = ast.parse(open('odoo/release.py', 'rb').read(), 'odoo/release.py')
assignment = next((n for n in ast.walk(mod) if isinstance(n, ast.Assign) if n.targets[0].id == 'version_info'), None)
version = assignment and '{0.n}.{1.n}'.format(*assignment.value.elts)
return version
```

runbot/container.py
```diff
@@ -0,0 +1,160 @@
...
+    _logger.addHandler(handler)
+    _logger.info('Start container tests')
+
+    if len(sys.argv) < 5:
+        _logger.error('Missing arguments: "%s build_dir odoo_version odoo_port db_name"', sys.argv[0])
+        sys.exit(1)
+    build_dir = sys.argv[1]
+    odoo_version = sys.argv[2]
+    odoo_port = sys.argv[3]
+    db_name = sys.argv[4]
```
xmo-odoo: Might as well use ArgumentParser at this point no?

runbot/container.py
```diff
@@ -0,0 +1,160 @@
...
+    _logger.info('Building docker image')
+    dbuild = subprocess.run(['docker', 'build', '--tag', 'odoo:runbot_tests'])
+
+if __name__ == '__main__':
+    _logger.setLevel(logging.DEBUG)
+    handler = logging.StreamHandler()
+    handler.setLevel(logging.DEBUG)
+    formatter = logging.Formatter('%(asctime)s %(levelname)s %(name)s: %(message)s')
+    handler.setFormatter(formatter)
+    _logger.addHandler(handler)
```
xmo-odoo: Wouldn't `logging.basicConfig(level=logging.DEBUG, format='%(asctime)s %(levelname)s %(name)s: %(message)s')` do the trick?

PR #58, opened 2018-11-30
[IMP] runbot: add basic tests

d-fence:
Deploying non tested code is prone to errors and leads to frustration.
With this commit, basic tests are done to avoid such a situation.

### DIFF
--- a/runbot/tests/test_frontend.py
+++ b/runbot/tests/test_frontend.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+from itertools import cycle
+from unittest.mock import patch
+from werkzeug.wrappers import Response
+from odoo.tests import common
+from odoo.addons.runbot.controllers import frontend
+
+
+class Test_Frontend(common.HttpCase):
+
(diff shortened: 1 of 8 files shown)


runbot/tests/test_frontend.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertEqual(template, 'runbot.repo', 'The frontend controller should use "runbot.repo" template')
+        # return the context to verify it in the tests
+        return context  # should find something smart to avoid the warning
+
+    @patch('odoo.addons.runbot.controllers.frontend.request')
+    def test_frontend_basic(self, mock_request):
+        mock_request.env = self.env
+        mock_request.render = self.mocked_render
+        mock_request._cr = self.cr
+        controller = frontend.Runbot()
```
d-fence: @xmo-odoo not sure that it's the right way to test a controller as the mocked_render generates a warning because it returns an invalid response for an http request.
Maybe I could use a closure and put the assertions inside it and finally try to return a valid response.
Any idea ?

PR #59, opened 2018-12-05
[FIX] runbot: reduce the amount of github status

d-fence:
When a commit is built serveral times, each time the status is updated, a
request is sent to github for each build.
As a side effect, if the first build is a failure and the last one a
success, github is wrongly updated to failure.
This bug is particulary annoying on PR in conjunction with the
runbot_merge, preventing the PR to be merged even after several
rebuild.

With this commit, only the latest build per repository is used to update
the github status. The amount of github status updates is also reduced.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -722,27 +722,28 @@ class runbot_build(models.Model):
         """Notify github of failed/successful builds"""
         runbot_domain = self.env['runbot.repo']._domain()
         for build in self:
-            b = build.duplicate_id if build.state == 'duplicate' else build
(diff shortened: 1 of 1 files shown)


runbot/models/build.py
```diff
@@ -838,8 +838,11 @@ def _job_30_run(self, build, lock_path, log_path):
             v['result'] = "ko"
         build.write(v)
         # post statuses on duplicate too
-        for b in self.search([('name', '=', build.name)]):
```
d-fence: self.search([('name', '=', build.name)])._github_status()

PR #60, opened 2018-12-06
[FIX] runbot, runbot_cla: notify each repo of cla status

d-fence:
When a PR is a duplicate of a branch, only the branch CLA status are
update. The same issue for build status was fixed in commit 4f1a55da9.

With this commit, there is new method than can be used in runbot_cla.
This method updates commit given status in each repo.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -718,6 +718,14 @@ class runbot_build(models.Model):
         p = subprocess.Popen(cmd, stdout=out, stderr=out, preexec_fn=preexec_fn, shell=shell, env=env, close_fds=False)
         return p.pid
 
+    def _github_status_notify_all(self, status):
+        """Notify each repo with a status"""
+        self.ensure_one()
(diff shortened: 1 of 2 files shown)


runbot/models/build.py
```diff
@@ -718,6 +718,14 @@ def preexec_fn():
         p = subprocess.Popen(cmd, stdout=out, stderr=out, preexec_fn=preexec_fn, shell=shell, env=env, close_fds=False)
         return p.pid
 
+    def _github_status_notify_all(self, status):
+        """Notify each repo with a status"""
+        self.ensure_one()
+        commits = {(b.repo_id, b.name) for b in self.search([('name', '=', self.name)])}
+        for repo, commit_hash in commits:
+            _logger.debug("github updating status %s to %s in repo %s", commit_hash, status['state'], repo.name)
```
xmo-odoo: I'd use `status['context']` instead of the status word or something, so it's clear in the log whether it's sending legal/cla or ci/runbot.

PR #61, opened 2018-12-12
[IMP] runbot_merge: send feedback when merge method is changed

d-fence:
When a user changes the merge method via github messages, no feedback is
sent. This could lead to strange behavior, for example when a user try
to joke with the mergebot like this:
robodoo are you goin ti merge my PR rogntudju !
This set the merge method to "merge" and the user is not aware of it.

With this commit, a feedback is sent via github.


xmo-odoo:
@d-fence works for me, though the tests were failing so I updated them, and I replaced the bare value by its label for (hopefully) better clarity.

PR #62, opened 2018-12-12
[FIX] runbot: fix name in branch matching for pull requests

ap-wtioit:
since 439e336a2f0e401b7cae6bb401d3ad77694db1a9 the pull_head_name
consists of the label (project_name:branch_name) therefore we need only
the last part of the label for finding the closest branch

Info @wt-io-it

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -186,8 +186,10 @@ class runbot_build(models.Model):
         )
         branches = sorted(branches, key=sort_by_repo)
 
+        # for prefix matching we remove the pull request prefix (project_name:) from the pull_head_name
+        name_for_prefix_match = branch.pull_head_name and branch.pull_head_name.split(':', 1)[-1] or branch.branch_name
         for branch in branches:
-            if name.startswith(branch['branch_name'] + '-') and self._branch_exists(branch['id']):
+            if name_for_prefix_match.startswith(branch['branch_name'] + '-') and self._branch_exists(branch['id']):
                 return result_for(branch, 'prefix')
 
         # 4. Common ancestors (git merge-base)


ap-wtioit:
@d-fence could you have a look at this

merge requests at our runbot always where build against default branches.

The same thing happened on the official runbot for http://runbot.odoo.com/runbot/build/413331 (https://github.com/odoo/enterprise/pull/2751) where runbot chose to checkout odoo master instead of saas-11.5

xmo-odoo:
@ap-wtioit I don't know that the change is correct: if there's a PR from "ap-wtioit:master" in enterprise and an other one in community, we want to build them against one another specifically, not build the first against community's own master.

ap-wtioit:
@xmo-odoo ok, i didn't have this use case previously, i will update the pull request to only affect the prefix code branch if no exactly matching branch is found

d-fence:
@ap-wtioit in the [example](https://github.com/odoo/enterprise/pull/2751) that you mention, the branch is badly named 'saas11.5' instead of 'saas-11.5', preventing the prefix branch matching to work.

ap-wtioit:
@d-fence didn't see that, thanks, then there is no example for the error on official runbot yet

d-fence:
I agree that there could still be errors in branch matching and I would be pleased to merge your PR if it does not break our instance :smile:

ap-wtioit:
@xmo-odoo @d-fence thanks for the feedback, i think i fixed it, it is now only touching the prefix match case

when https://github.com/odoo/runbot/pull/58 gets merged i'll have a look to add our test cases as well

mreficent:
@ap-wtioit as you said, https://github.com/odoo/runbot/pull/58 was merged and thus you can take a look

Xavier-Do:
Branch matching changed a lot since this pr and some corner cases where already fixed. Looks like it is outdated.

PR #63, opened 2018-12-18
Support for blacklisting branches (builds-wise)

xmo-odoo:
On our runbot, there's no reason to build or mount tmp branches (they're where runbot_merge sets up / creates the branches it's eventually staging so there's a lot of churn none of which matters).
Also, there is little reasons to keep staging branches running after the build.

This PR adds a selection field to choose the kind of jobs wanted for the branches, that way the tmp branches can be ignored and the staging branches can be built but not keeping running.

The ability to create ignore patterns at the repo level (or somesuch) might eventually be useful, but for now it looks way overkill: new tmp branches only get defined when adding new branches/targets to runbot_merge, and it's not super problematic if we forget to blacklist one such branch, it just creates some unnecessary builds/churns.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -25,6 +25,19 @@ re_job = re.compile('_job_\d')
 _logger = logging.getLogger(__name__)
 
 
+def runbot_job(*accepted_job_types):
+    """ Decorator for runbot_build _job_x methods to filter build jobs """
+    accepted_job_types += ('all', )
+
+    def job_decorator(func):
+        def wrapper(self, build, log_path):
+            if build.job_type == 'none' or build.job_type not in accepted_job_types:
+                build._log(func.__name__, 'Skipping job')
+                return -2
+            return func(self, build, log_path)
+        return wrapper
+    return job_decorator
+
 class runbot_build(models.Model):
 
     _name = "runbot.build"
@@ -74,6 +87,13 @@ class runbot_build(models.Model):
                                    ],
                                   default='normal',
                                   string='Build type')
+    job_type = fields.Selection([
+        ('testing', 'Testing jobs only'),
+        ('running', 'Running job only'),
+        ('all', 'All jobs'),
+        ('none', 'Do not execute jobs'),
+        ],
+    )
 
     def copy(self, values=None):
         raise UserError("Cannot duplicate build!")
@@ -81,6 +101,8 @@ class runbot_build(models.Model):
     def create(self, vals):
         build_id = super(runbot_build, self).create(vals)
         extra_info = {'sequence': build_id.id}
+        job_type = vals['job_type'] if 'job_type' in vals else build_id.branch_id.job_type
+        extra_info.update({'job_type': job_type})
         context = self.env.context
 
         # detect duplicate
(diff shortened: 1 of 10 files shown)


runbot/models/build.py
```diff
@@ -746,6 +755,9 @@ def _job_02_docker_build(self, build, log_path):
         return -2
 
     def _job_10_test_base(self, build, log_path):
+        if build.branch_id.job_type in ('none', 'running'):
```
KangOl: Wouldn't it be better to have a decorator to specify the job type?

```python
@runbot_job('testing')
def _job_10_test_base(self, build, log_path):
    ...
```

d-fence: It looks like a really good idea.
Anyway, this PR is not finished, I need to adapt to the build.job_type

runbot/models/build.py
```diff
@@ -732,19 +752,22 @@ def _github_status(self):
...
         return -2
 
+    @runbot_job('testing', 'running', 'all')
     def _job_02_docker_build(self, build, log_path):
         """Build the docker image"""
         build._log('docker_build', 'Building docker image')
         docker_build(log_path, build._path())
         return -2
 
+    @runbot_job('testing', 'all')
```
d-fence: @KangOl is it the kind of decorator that you were talking about

KangOl: Yes, but maybe you should handle the `all` flag in the decorator itself

runbot/models/build.py
```diff
@@ -782,15 +811,15 @@ def _job_20_test_all(self, build, log_path):
         build.write({'job_start': now()})
         return docker_run(cmd, log_path, build._path(), build._get_docker_name(), cpu_limit=cpu_limit)
 
+    @runbot_job('testing')
     def _job_21_coverage_html(self, build, log_path):
-        if not build.coverage:
-            return -2
```
xmo-odoo: Doesn't that mean coverage will run on all testing jobs, not just builds/branches flagged as coverage?

d-fence: well catch, I removed it by accident. Thanks.

runbot_cla/runbot.py
```diff
@@ -13,7 +14,10 @@
 class runbot_build(models.Model):
     _inherit = "runbot.build"
 
+    @runbot_job('testing')
     def _job_05_check_cla(self, build, log_path):
+        if job_type == 'none':
+            return -2
```
xmo-odoo: Doesn't `@runbot_job` already skip jobs when their type is none? Also where does the job_type local come from?

d-fence: sure

runbot/models/build.py
```diff
@@ -25,6 +25,19 @@
 _logger = logging.getLogger(__name__)
 
 
+def runbot_job(*accepted_job_types):
+    """ Decorator for runbot_build _job_x methods to filter build jobs """
+    accepted_job_types += ('all', )
+
+    def job_decorator(func):
+        def wrapper(self, build, log_path):
+            if build.job_type == 'none' or build.job_type not in accepted_job_types:
```
xmo-odoo: Alternative: special case `all` instead of `none`? e.g. `if build.job_type != 'all' and job_type not in accepted_job_types: skip()`, and remove line 30?

PR #65, opened 2019-01-03
[FIX] runbot: speedup and limit search in frontend

d-fence:
When searching the builds for the frontend the resulting query can last
a very long time (up to 7sec).

With this commit, the search result is strictly limited to 1000 builds
and the search string length is limited to 60 chars.

Also, the branch_id many2one field on build is set to auto_join in order
to generate a better sql query. Thanks @tbe-odoo.

### DIFF
--- a/runbot/controllers/frontend.py
+++ b/runbot/controllers/frontend.py
@@ -47,7 +47,8 @@ class Runbot(Controller):
         return count, level
 
     @route(['/runbot', '/runbot/repo/<model("runbot.repo"):repo>'], website=True, auth='public', type='http')
-    def repo(self, repo=None, search='', limit='100', refresh='', **kwargs):
+    def repo(self, repo=None, search='', refresh='', **kwargs):
+        search = search if len(search) < 60 else search[:60]
         branch_obj = request.env['runbot.branch']
         build_obj = request.env['runbot.build']
         repo_obj = request.env['runbot.repo']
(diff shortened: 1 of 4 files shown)


runbot/controllers/frontend.py
```diff
@@ -49,6 +49,8 @@ def _pending(self):
 
     @http.route(['/runbot', '/runbot/repo/<model("runbot.repo"):repo>'], website=True, auth='public', type='http')
     def repo(self, repo=None, search='', limit='100', refresh='', **kwargs):
+        limit = limit if int(limit) <= 1000 else '1000'
+        search = search if len(search) < 60 else search[:60]
```
julienlegros: Setting a maximum for `limit` is definitely a good idea, but I'm not sure limiting the search length is really useful.

d-fence: On the other hand, do you think that searching more than 60 chars is a real use case ?
This might discard automated scripted long queries, they will not even reach the database engine.

KangOl:
The first `WHEN` in `_guess_result()` `CASE` can be removed

tbe-odoo:
Hello guys,

There is a bug in the ORM where fields in `auto_join` cannot be used on domains that have OR conditions, so I am not sure if you should keep the `auto_join` :/

https://github.com/odoo/odoo/issues/25175

PR #66, opened 2019-01-07
[FIX] runbot: move transifex config dir to odoo/addons

d-fence:
The odoo transifex module uses the .tx/config file to find translated
modules. This module search for .tx/config file in the addons_path
and their parent directories. In the case of an enterprise build, the
repo is exported to the build directory and right after that, the
community (server) repo is extracted, the .tx/config from enterprise is
overwritten.

With this commit, the .tx directory is moved to odoo/addons before it
gets overwritten by the one from the server.


d-fence:
renamed the branch to remove the typo

PR #77, opened 2019-01-22
[FIX] runbot: write build results in a separate job

d-fence:
When a build reach the job30_run method, results from a previous testing
methods are computed.
With the previous commit 8c73e6a90133c8915ced619a31e48b10fb01b70d this
job can now be skipped. In that case, the results are not set.

With this commit, the results are computed in a separate method.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -836,10 +836,9 @@ class runbot_build(models.Model):
             build._log('coverage_result', 'Coverage file not found')
         return -2  # nothing to wait for
 
-    @runbot_job('running')
-    def _job_30_run(self, build, log_path):
-        # adjust job_end to record an accurate job_20 job_time
-        build._log('run', 'Start running build %s' % build.dest)
+    @runbot_job('testing', 'running')
+    def _job_29_results(self, build, log_path):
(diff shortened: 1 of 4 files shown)


runbot/models/build.py
```diff
@@ -856,6 +855,12 @@ def _job_30_run(self, build, log_path):
             v['result'] = "ko"
         build.write(v)
         build._github_status()
+        return -2
```
xmo-odoo: Maybe this magic value should be a documented constant instead of just a magic number?

In fact, in a post-dockerisation system I don't think storing the PID (and returning it from job methods) is even necessary anymore, jobs could just return e.g. True/False or SOME_VALUE / None (not return anything) to say whether they've launched a sub-process (or not in which case a new scheduling should be performed immediately): the container name is used to kill processes (which was previously done through the pid, and I guess is why the default value would be -2: 0 and -1 are both magic pids, -2 is not).

d-fence: I agree, but I kept it to ease migration. The first idea was to store the container name instead ans some magic values for jobs not spawning containers.
Anyway, it could be done in another PR/task/branch.

PR #86, opened 2019-01-31
[IMP] runbot: add a small test for _local_pg methods

d-fence:
Small test to prevent annoying deployment

### DIFF
--- a/runbot/tests/test_build.py
+++ b/runbot/tests/test_build.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
+import psycopg2
+import uuid
 from unittest.mock import patch
-from odoo.tools.config import configmanager
 from odoo.tests import common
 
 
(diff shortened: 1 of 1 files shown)


d-fence:
@KangOl any idea of an assert that could check that the db is created/dropped ?

KangOl:
You should at least connect to the database and make a query (leaving the cursor open)

To ensure the database is gone, you can search for it in `pg_database` table (with local cursor)

I would also use a random database name like `str(uuid.uuid4())`

runbot/tests/test_build.py
```diff
@@ -58,6 +59,24 @@ def test_build_cmd_log_db(self, mock_grep, mock_mkdir):
...
 
+    def test_build_local_pg(self):
+        """ test create and drop of a local database even with an open cursor"""
+        dbname = str(uuid.uuid4())
+        self.Build._local_pg_createdb(dbname)
+        cnx = psycopg2.connect("dbname=%s" % dbname)
+        cur = cnx.cursor()
+        cur.execute("CREATE TABLE test (id serial PRIMARY KEY, foo varchar);")
+        cur.execute("INSERT INTO test (foo) VALUES ('bar')")
+        cnx.commit()
```
KangOl: Isn't it better to keep the cursor open?

Xavier-Do:
Creating a database in a test is a bad idea, this should be at least -standard. Anyway, this was open for a while without being merged, can we close this? :)

PR #91, opened 2019-02-06
[REF] runbot: split cron to speed up builds

d-fence:
Then Runbot Cron is executed on each runbot instance. When the number of
instances scales, the time needed for an instance to obtain the cron
increases.

With this commit, the Runbot cron only executes the _update mtehod.
That way, only one runbot instance fetch the respositories and create
pending builds.

On the other hand, a dedicated cron have to be manually created for each
runbot instance. Those cron's only have to call the _cron_by_host
method, passing the hostname as parameter. This method will then self
assign pending builds if there are slots available. In that case, if the
build hash cannot be found in the local repo, the repo is fetched.


runbot/models/build.py
```diff
@@ -440,6 +440,11 @@ def _schedule(self):
                     'job_end': False,
                 }
                 build.write(values)
+                # as the reporistory update is now in an independant cron, we feth if needed
+                try:
+                    build.repo_id._git(['show', '--oneline', build.name])
```
xmo-odoo: `—oneline` seems to do the opposite of what we want on `git show`.

Also in any case git show seems complex and inefficient, `cat-file -e` (or `cat-file -t` and checking that it prints "commit" to stdout) seems like a better idea.

runbot/models/repo.py
```diff
@@ -262,10 +262,16 @@ def _scheduler(self, ids=None):
...
 
             # find sticky / priority pending build if any, otherwise, last pending (by id, not by sequence) will do the job
-
-            pending_ids = Build.search(domain + [('state', '=', 'pending'), '|', ('branch_id.sticky', '=', True), ('branch_id.priority', '=', True)], limit=1)
-            if not pending_ids:
-                pending_ids = Build.search(domain + [('state', '=', 'pending')], order="sequence", limit=1)
+            # obtain a lock on the first pending build to be sure that another runbot instance will not schedule it
+            query = """"SELECT runbot_build.id FROM runbot_build
+                        LEFT JOIN runbot_branch ON runbot_branch.id = runbot_build.branch_id
+                        WHERE runbot_build.repo_id IN %(repo_ids)s
```
xmo-odoo: `runbot_branch.job_type` is missing from the filtering but I don't know if that's a problem.

runbot/models/repo.py
```diff
@@ -262,10 +262,16 @@ def _scheduler(self, ids=None):
...
+                        AND runbot_build.state='pending'
+                        ORDER BY runbot_branch.sticky DESC, runbot_branch.priority DESC, runbot_build.sequence ASC
+                        LIMIT 1
+                        FOR UPDATE OF runbot_build SKIP LOCKED"""
+            self.env.cr.execute(query, {'repo_ids': ids})
```
xmo-odoo: Are we sure `ids` is always a tuple? `in` breaks if given a list (~array) (or an empty tuple but I assume `ids` is never empty).

d-fence: I searched builds without a repo_id and didn't find one :smile:

runbot/models/build.py
d-fence: Was not aware of git cat-file, learning something each day is great

runbot/models/repo.py
d-fence: Of course, you're right

runbot/models/repo.py
```diff
@@ -324,5 +330,13 @@ def _reload_nginx(self):
     def _cron(self):
         repos = self.search([('mode', '!=', 'disabled')])
         self._update(repos)
+
+    def _cron_by_host(self, hostname):
```
KangOl: it's actually `_cron_for_host()` no?

d-fence: Ok, I change.

runbot/models/repo.py
KangOl: It's the other way around.
Calling `_scheduler()` without `ids`. Thus not filtering on repo.

d-fence: Do you mean that the SQL query could fail ?

d-fence: no ids means no enabled repository. So better to return early from _scheduler in that particular case

KangOl:
It look like monkey-patching `_process_job` is not needed.
If `_handle_callback_exception` reraise the exception, the update of `nextcall` won't happen.
```python
class ir_cron(odoo.modes.Model):
    _inherit = 'ir.cron'

    @api.model
    def _handle_callback_exception(self, cron_name, server_action_id, job_id, job_exception):
       super()._handle_callback_exception(cron_name, server_action_id, job_id, job_exception)
       if isinstance(job_exception, CronHostError):
           raise job_exception
```

xmo-odoo:
@KangOl one issue in that case is there will be a ton of tracebacks in the logs, especially as CronHostError will be pretty much routine and expected. Worse, they'll be doubled as both `_callback` and `_process_jobs` use logging.exception.

d-fence:
I agree, at the beginning I was logging a small line to say 'skipped blah blah' but even that is too much because 11 runbot / 10 sec makes a lot of noise.

KangOl:
We can just filter out the logging then.

```python
def no_log_CronHostError(record):
    if record.exc_info and record.exc_info[0] is CronHostError:
        return False
    return True

logging.getLogger('odoo.addons.base.models.ir_cron').addFilter(no_log_CronHostError)
```

runbot/models/repo.py
```diff
@@ -256,22 +272,27 @@ def _scheduler(self, ids=None):
...
+        available_slots = workers - nb_testing
+        if available_slots > 0:
+            # self-assign to be sure that another runbot instance cannot self assign the same builds
+            query = """UPDATE runbot_build
+                       SET host=%(host)s
+                       WHERE runbot_build.id in (
+                         SELECT runbot_build.id FROM runbot_build
+                         LEFT JOIN runbot_branch ON runbot_branch.id = runbot_build.branch_id
+                         WHERE runbot_build.repo_id IN %(repo_ids)s
+                         AND runbot_build.state='pending'
```
odony: lint/prettify the SQL query a bit? e.g. consistent spaces, etc.?

runbot/models/repo.py
```diff
@@ -256,22 +272,27 @@ def _scheduler(self, ids=None):
...
+                         AND runbot_branch.job_type != 'none'
+                         AND runbot_build.host is null
+                         ORDER BY runbot_branch.sticky DESC, runbot_branch.priority DESC, runbot_build.sequence ASC
+                         LIMIT %(available_slots)s
```
odony: I would put the LIMIT clause last in the query, to better show that it is applied *after* skipping any locked rows ;-)

KangOl:
Now, with long-running crons, we don't need `CronHostError` anymore. Simply return when host doesn't match

runbot/models/repo.py
```diff
@@ -342,19 +353,40 @@ def _reload_nginx(self):
...
+
+    def _cron_fetch_and_schedule(self, hostname):
+        """This method have to be called from a dedicated cron on a runbot
+        in charge of orchestration.
+        """
+        if hostname != fqdn():
+            raise CronHostError('Not for me')
+        start_time = time.time()
+        timeout = self._get_cron_period()
+        while time.time() - start_time > timeout:
```
xmo-odoo: Shouldn't that be `<`?

Also why not `end_time = time.time() + self._get_cron_period()` and loop while `time.time() < end_time`?

d-fence: :+1:  :see_no_evil:

xmo-odoo:
There's one thing which really bothers me about the "long cron" concept: it's a single transaction which might last a long, long time. That seems really fucky. In fact it's *two* transactions lasting a long long time, the cron's and the job's.

KangOl:
You're right.
The job cursor should be committed at each loop.

odony:
> There's one thing which really bothers me about the "long cron" concept: it's a single transaction which might last a long, long time. In fact it's _two_ transactions lasting a long long time, the cron's and the job's.

Well, it's really the way most of our "queue runners" jobs are meant to be designed. The cron transaction only serves as a semaphore, while the main job is meant to execute many small transactions, ensuring maximum throughput with minimal risk of side-effect / duplicate work in case of interruption/rollback.
Mails, recurring payments, procurements, ... are all processed by cron jobs in this manner.

In fact, this is precisely why we [introduced 2 distinct cursors in 7.0](https://github.com/odoo/odoo/commit/75c9cf255a6d64f47e91798f1d89379da72d5886) in the first place: the job cursor is meant to be frequently committed, so it cannot be the one holding the cron lock.

Apart from rollbacks, which are of little concern here (by construction), what problem do you foresee with long-running transactions?

runbot/models/repo.py
```diff
@@ -256,22 +267,37 @@ def _scheduler(self, ids=None):
...
-            pending_ids._schedule()
-
-            # compute the number of testing and pending jobs again
-            testing = Build.search_count(domain_host + [('state', '=', 'testing')])
-            pending = Build.search_count(domain + [('state', '=', 'pending')])
+        nb_testing = Build.search_count(domain_host + [('state', '=', 'testing')])
+        available_slots = workers - nb_testing
+        if available_slots > 0:
+            # self-assign to be sure that another runbot instance cannot self assign the same builds
+            query = """UPDATE
```
odony: commit just before the select for update

PR #92, opened 2019-02-14
[FIX] runbot: search common ancestor against build branch

KangOl:
4 year bug introduction by 2e01744b0df61e6ab123c3b99b6bc90f18549652 in
which, the variable `branch` has been shadowed by the for loop in case
3.

42ae78f6476c7c12144caf33f656fd22b72299dc try to fix it, but incorrectly.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -230,7 +230,7 @@ class runbot_build(models.Model):
             """, [repo.id, target_id])
             for common_name, in self.env.cr.fetchall():
                 try:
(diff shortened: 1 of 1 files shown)


xmo-odoo:
@d-fence wasn't that already fixed separately?

KangOl:
"fixed" by removing the feature...

d-fence:
feature removed as it never worked without anyone complaining.
al confirmed that it was to be removed

PR #95, opened 2019-02-21
[FIX] runbot: limit number of log lines shown on build page

d-fence:
When displaying build logs, all the messages from ir_logging about this
particular build are fetched from the database.

From time to times, it happens that the number of logged messages is
really huge. Those messages lines could also contain multiple lines,
multiplying the number of row to generate in the html page.
When this happens, the process that generates the template last a long
time and ends with a MemoryError. If the end user, bored, hits the
refresh button multiple times, all the workers will be busy building
this template. In the end, all users get a Bad Gateway from nginx.

With this commit, the number of messages that will be taken into account
will be limited to 10000.

### DIFF
--- a/runbot/controllers/frontend.py
+++ b/runbot/controllers/frontend.py
@@ -190,7 +190,7 @@ class Runbot(Controller):
             domain.append(('level', '=', level.upper()))
         if search:
             domain.append(('message', 'ilike', search))
-        logging_ids = Logging.sudo().search(domain)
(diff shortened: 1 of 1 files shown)


d-fence:
@julienlegros the avg number of messages is very small (around 250) and the peak causing the error is approximately between 40000 and 50000. Il will discuss the idea of removing the running logs completely.

PR #96, opened 2019-02-25
[FIX] runbot: update repo if needed before exporting

d-fence:
When a runbot execute the cron_fetch_and_build method, a new pending
build can be created between the repo update and the scheduling.
If it happens, the runbot tries to export a hash reference the does not
exists yet.

With this commit, the repo update is done right before exporting the
repo and only if the commit hash is not found.

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -90,6 +90,15 @@ class runbot_repo(models.Model):
         p1.stdout.close()  # Allow p1 to receive a SIGPIPE if p2 exits.
         p2.communicate()[0]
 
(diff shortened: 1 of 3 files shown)


runbot/models/repo.py
```diff
@@ -90,6 +90,15 @@ def _git_export(self, treeish, dest):
         p1.stdout.close()  # Allow p1 to receive a SIGPIPE if p2 exits.
         p2.communicate()[0]
 
+    def _hash_exists(self, hash):
```
xmo-odoo: hash is a builtin function so probably name the parameter something else.

PR #103, opened 2019-03-05
Add sentry integration to more easily track issues

xmo-odoo:
Currently, we track mergebot (& runbot?) issues by checking the logs once in a while, or when colleagues arrive screaming the end of the world 
philippulus_the_prophet

Sending warnings and errors to sentry seems like it could be useful there. The saas already has an integration, but it uses the old client (raven as opposed to python-sdk) and it seems a bit extreme to install saas-management modules to get sentry reporting when the integration looks so simple (though I could have missed stuff).

OTOH this is not quite ideal as our org is permanently rate-limited. Still, issues do seem to go through (testing locally with reportable warning) so…

### DIFF
--- a/sentry/__init__.py
+++ b/sentry/__init__.py
@@ -1,0 +1,20 @@
+import logging
+from os import environ
+
+import sentry_sdk
+from sentry_sdk.integrations.wsgi import SentryWsgiMiddleware
+from sentry_sdk.integrations.logging import LoggingIntegration
+from odoo import http
+
+def load_sentry():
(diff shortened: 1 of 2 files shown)


xmo-odoo:
@julienlegros suggested just doing it in mergebot until and unless runbot wants to eventually use it, as it's a bit odd having a "sentry" module in the runbot repo, but at the same time there isn't really a good home for it (having its own repo for 10 lines of code would be excessive, and odoo-extra might work but I don't think it's normally deployed on our machines).

PR #112, opened 2019-04-04
[IMP] runbot: detach odoo command from docker_run

d-fence:
When the docker_run function is called, the odoo command is decorated
with a pip command to install required packages.
This pollute the docker_run function if a runbot job_ method wants to
use docker for something else that starting an odoo instance (like
pg_dump) for example.

With this commit, command modification is made in an optional helper
function named build_odoo_cmd.

the docker_run function now needs the command to run as a string instead
of a list of odoo cmd and its parameters.

### DIFF
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -30,6 +30,20 @@ USER odoo
 ENV COVERAGE_FILE /data/build/.coverage
 """ % {'group_id': os.getgid(), 'user_id': os.getuid()}
 
+
+def build_odoo_cmd(odoo_cmd):
+    """ returns the chain of commands necessary to run odoo inside the container
+        : param odoo_cmd: odoo command as a list
+        : returns: a string with the command chain to execute in the docker container
+    """
+    # build cmd
+    cmd_chain = []
+    cmd_chain.append('cd /data/build')
+    cmd_chain.append('head -1 odoo-bin | grep -q python3 && sudo pip3 install -r requirements.txt || sudo pip install -r requirements.txt')
+    cmd_chain.append(' '.join(odoo_cmd))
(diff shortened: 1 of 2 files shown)


d-fence:
As a side note:
I encountered a problem while testing this PR. When the I use the "psql" command instead of odoo-bin like [here](https://github.com/odoo/runbot/pull/112/files#diff-a781692ae7da61310cb61986595e67a2R172), if the postgres version on the host is different from container, the command fails. But it's not directly linked to the PR code.

runbot/container.py
```diff
@@ -153,13 +161,24 @@ def tests(args):
...
+    if args.dump:
+        _logger.info("Testing pg_dump")
+        logfile = os.path.join(args.build_dir, 'logs', 'logs-pg_dump.txt')
+        container_name = 'odoo-container-test-pg_dump-%s' % datetime.datetime.now().microsecond
+        docker_pg_dump_cmd = 'cd /data/build/datadir && pg_dump -U %s -f db_export.sql %s' % (os.getlogin(), args.db_name)
+        docker_run(docker_pg_dump_cmd, logfile, args.build_dir, container_name)
+        time.sleep(1)
+        while docker_is_running(container_name):
+            time.sleep(10)
+            _logger.info("Waiting for %s to stop", container_name)
```
xmo-odoo: Maybe these wait loops could just use [docker wait](https://docs.docker.com/engine/reference/commandline/wait/) instead?

d-fence: Great, I was not aware of this docker command. Thanks

d-fence: On the other hand, it tests the "docker_is_running" function

PR #118, opened 2019-04-19
[FIX] runbot: use a repo short name to avoid bugs in qweb template

d-fence:
When using a local git repo, the git name does not have colon, making
the frontend crash.

With this commit, a non-stored computed field 'short_name' is added to
compute a shortest version of the name.

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -24,7 +24,7 @@ class runbot_repo(models.Model):
     _name = "runbot.repo"
 
     name = fields.Char('Repository', required=True)
-    # branch_ids = fields.One2many('runbot.branch', inverse_name='repo_id') # keep for next version
+    short_name = fields.Char('Repository', compute='_compute_short_name', store=False, readonly=True)
     sequence = fields.Integer('Sequence')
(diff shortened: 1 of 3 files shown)


runbot/models/repo.py
```diff
@@ -74,6 +74,11 @@ def _get_base_url(self):
             name = name.replace(':', '/')
             repo.base = name
 
+    @api.depends('name', 'base')
+    def _compute_short_name(self):
+        for repo in self:
+            repo.pretty_name = '/'.join(repo.base.split('/')[-2:])
```
Xavier-Do: Should't we split on ':' to? I think that the base format we have on runbot is ...:odoo/odoo.

d-fence: That's why I use repo.base instead of repo.name, the colon is replaced by a '/'

https://github.com/odoo/runbot/blob/82f881d9e6d6d72fa4878a23c770a5a92f26bb77/runbot/models/repo.py#L74

Xavier-Do: Ho ok great, in this case looks good to me!

PR #122, opened 2019-04-29
When staging, wait for the new head to be visible at the repo-level before staging the next repo

xmo-odoo:
Replace the arbitrary sleep after having staged a head by checking for the head to be visible to regular git client through the HTTP v1 "smart" protocol: `$GIT_URL/info/refs?service=git-upload-pack` returns a listing of all the repo's refs and should be what regular git clients see (and the first thing they fetch) when fetching a repo.

Also alter the semantics slightly: if a head is never seen, the staging now fails (rather than carrying on blindly with staging the dependents), though it's not rolled back (maybe it should be?)

Performs one check immediately, because for smaller repositories the update is essentially instantaneous.

### DIFF
--- a/runbot_merge/models/pull_requests.py
+++ b/runbot_merge/models/pull_requests.py
@@ -10,13 +10,15 @@ import time
 
 from itertools import takewhile
 
+import requests
+
 from odoo import api, fields, models, tools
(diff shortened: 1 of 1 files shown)


d-fence:
:+1:

PR #124, opened 2019-04-30
[IMP] runbot: split branch creation and pending builds creation

d-fence:
When searching for new builds by parsing git refs, the new branches are
created as well as the pending builds in the same _find_new_commits
method.

With this commit, this behavior is splitted into two methods, that way,
it's now possible to create missing branches without creating new
builds.

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -133,21 +133,26 @@ class runbot_repo(models.Model):
                 else:
                     raise
 
-    def _find_new_commits(self):
-        """ Find new commits in bare repo """
+    def _get_refs(self):
+        """Find new refs
+        :return: list of tuples with following refs informations:
+        name, sha, date, author, author_email, subject, committer, committer_email
+        """
         self.ensure_one()
-        Branch = self.env['runbot.branch']
-        Build = self.env['runbot.build']
-        icp = self.env['ir.config_parameter']
-        max_age = int(icp.get_param('runbot.runbot_max_age', default=30))
-
         fields = ['refname', 'objectname', 'committerdate:iso8601', 'authorname', 'authoremail', 'subject', 'committername', 'committeremail']
         fmt = "%00".join(["%(" + field + ")" for field in fields])
         git_refs = self._git(['for-each-ref', '--format', fmt, '--sort=-committerdate', 'refs/heads', 'refs/pull'])
         git_refs = git_refs.strip()
+        return [tuple(field for field in line.split('\x00')) for line in git_refs.split('\n')]
 
-        refs = [[field for field in line.split('\x00')] for line in git_refs.split('\n')]
-
+    def _find_or_create_branches(self, refs):
+        """Parse refs and create branches that does not exists yet
+        :param refs: list of tuples returned by _get_refs()
+        :return: dict {branch.name: branch.id}
+        The returned structure contains all the branches from refs newly created
+        or older ones.
+        """
+        Branch = self.env['runbot.branch']
         self.env.cr.execute("""
             WITH t (branch) AS (SELECT unnest(%s))
           SELECT t.branch, b.id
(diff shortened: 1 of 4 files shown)


runbot/models/repo.py
```diff
@@ -158,12 +160,22 @@ def _find_new_commits(self):
...
+        :param refs: list of tuples returned by _get_refs()
+        """
+        self.ensure_one()
+        Branch = self.env['runbot.branch']
+        Build = self.env['runbot.build']
+        icp = self.env['ir.config_parameter']
+        max_age = int(icp.get_param('runbot.runbot_max_age', default=30))
+
+        for name, sha, date, author, author_email, subject, committer, committer_email in refs:
+            branch = Branch.search([('name', '=', name), ('repo_id', '=', self.id)], limit=1)
```
Xavier-Do: Here the search is in a loop wich is to avoid as much as possible. In the _create_branches, we have this information in the ref_branches[name]. 

I would rename _create_branches in _find_or_create_branches, wich returns the ref_branches. When a branch is creates, the new id is added to ref_branches. 
```python3
    new_branch = Branch.create({'repo_id': self.id, 'name': name}).id
    ref_branches[name] = new_branch.id
....
return ref_branches
```
Then in create_pending_builds, get the result and give it to _find_new_commits
```python3
         for repo in self:
            ref_branches = repo._create_branches(refs[repo])
            repo._find_new_commits(refs[repo], ref_branches)
```
And here, we can just use the already known refs.
```python3
def _find_new_commits(self, refs, ref_branches ):
....
    for name, sha, date, author, author_email, subject, committer, committer_email in refs:
        branch = ref_branches['name']
```

runbot/models/repo.py
```diff
@@ -220,13 +232,19 @@ def _find_new_commits(self):
...
+        for repo in self:
             try:
-                repo._find_new_commits()
+                refs[repo] = repo._get_refs()
             except Exception:
-                _logger.exception('Fail to find new commits in repo %s', repo.name)
+                _logger.exception('Fail to get refs for repo %s', repo.name)
+
+        for repo in self:
+            repo._create_branches(refs[repo])
```
Xavier-Do: Before, this part was in the try catch, now we are not sure that repo is in refs if _get_refs failed for one reason or another. I would add a 'if repo in refs' guard in case something failed with one repo to avoid an error here. 

Another solution is ti iterate on refs keys: for repo in refs:

Xavier-Do:
Looks good to me exept for the little comments. 
I would add in the commit message that now the branches are created for all repo before the builds. This will help for closest branch detection if pr and branches are detected in the same update.

runbot/models/repo.py
Xavier-Do: Some side notes, we could browse all branches as part of the same browse record. I need to check this with @rco-odoo  but I think that with the previous solution and this one, branches won't be prefetched togheter and a read will be performed at each loop turn (accessing branch.sticky). 

To check again, but the idea would be to return records in _create_branches
Here are some ideas

```python3
branches = self.env['runbot.branch'].browse(ref_branches.values())
for name, branch_id in ref_branches.items():
    for branch in branches:
        if branch_id = branch.id:
            break
    ref_branches[name] = branch
return ref_branches
```
but this one is o(n²) and not very sexy. (It must be possible to do it another way, in any case it is acceptable I think

```python3
branches = self.env['runbot.branch'].browse(ref_branches.values())
for branch in branches.items():
    ref_branches[branch.name] = branch
return ref_branches
```
This one is prefetching record by accessing names. Not a big deal since we will need to access them if we have new builds but we could do that in the _find_new_commits instead in this case.


Side note 2: could be interesting to use a read_group for builds.

runbot/models/repo.py
```diff
@@ -133,21 +133,23 @@ def _github(self, url, payload=None, ignore_errors=False):
...
+        """
         self.ensure_one()
-        Branch = self.env['runbot.branch']
-        Build = self.env['runbot.build']
-        icp = self.env['ir.config_parameter']
-        max_age = int(icp.get_param('runbot.runbot_max_age', default=30))
-
         fields = ['refname', 'objectname', 'committerdate:iso8601', 'authorname', 'authoremail', 'subject', 'committername', 'committeremail']
         fmt = "%00".join(["%(" + field + ")" for field in fields])
         git_refs = self._git(['for-each-ref', '--format', fmt, '--sort=-committerdate', 'refs/heads', 'refs/pull'])
```
Xavier-Do: I'm currious to know how many git_refs we have in prod. Looks like we willl have a tons of them no? We could maybe add a limit with --count, sorting them by date, but I'm not sure this is correct.

runbot/models/repo.py
```diff
@@ -174,8 +174,11 @@ def _find_new_commits(self, refs):
         icp = self.env['ir.config_parameter']
         max_age = int(icp.get_param('runbot.runbot_max_age', default=30))
 
+        refs_branch_names = list({r[0] for r in refs})
+        branches = Branch.search([('name', 'in', refs_branch_names), ('repo_id', '=', self.id)])
+
         for name, sha, date, author, author_email, subject, committer, committer_email in refs:
-            branch = Branch.search([('name', '=', name), ('repo_id', '=', self.id)], limit=1)
+            branch = branches.filtered(lambda rec: rec.name == name)
```
d-fence: @Xavier-Do I think that this refactor is simpler to avoid the search in the loop.
What do you think ?

runbot/models/repo.py
d-fence: 8542 in -dev
It takes ≈ 202 ms to list them.
 I'm not sure that it needs limiting for such a small number

runbot/models/repo.py
```diff
@@ -220,13 +235,20 @@ def _find_new_commits(self):
...
-                repo._find_new_commits()
+                refs[repo] = repo._get_refs()
             except Exception:
-                _logger.exception('Fail to find new commits in repo %s', repo.name)
+                _logger.exception('Fail to get refs for repo %s', repo.name)
+
+        for repo in self:
+            if repo in refs:
+                repo._create_branches(refs[repo])
+                repo._find_new_commits(refs[repo])
```
Xavier-Do: I think that _find_new_commits should be in another loop (or what I said about build creation improvement is no correct), and are you going to add a guard on build creation?

runbot/models/repo.py
Xavier-Do: Looks good but the filter in the loop is maybe not the best idea (same as the o(n²) solution I proposed.) We can check and test that tomorrow difficult to say without testing ;).

runbot/models/repo.py
d-fence: 21000 PR in odoo

PR #133, opened 2019-05-24
[IMP] runbot: grant rights for the build logs

d-fence:
When an Odoo instance is launched buy a build, the --log-db parameter is
used to write its logs into the runbot ir_logging table.
Also, a trigger on the postgresql server is used to add the build id on
the ir_logging.

Because of that, the user used in the --log-db parameter must have the
rights to do those operations. Before this commit, it was up to the
runbot administrator to properly setup thos rights.

With this commit, the rights are automatically granted when the logdb
uri is set or changed in the runbot settings.


d-fence:
@Xavier-Do I think that this one is ready

Xavier-Do:
This is not useful since local log were introduced.

PR #136, opened 2019-06-04
[IMP] runbot: display config name

Xavier-Do:


Xavier-Do:
Note, need to add build config in branch view, maybe in build view too

PR #137, opened 2019-06-05
[FIX] runbot: try to update github status multiple times

d-fence:
When updating github statuses, it happens that we face a "Bad gateway"
from github. In that case, the error is logged in the runbot logs and
that's it. As a consequence, when the runbot_merge is waiting status for
the staging branch and this kind of error occurs, the runbot_merge
timeouts and the users vainly search the reason.

With this commit, the runbot tries to update the status at least twice.
If it fails, an INFO message is logged on the build itself.

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -123,31 +123,39 @@ class runbot_repo(models.Model):
             return False
         return True
 
-    def _github(self, url, payload=None, ignore_errors=False):
+    def _github(self, url, payload=None, ignore_errors=False, nb_tries=2):
         """Return a http request to be sent to github"""
         for repo in self:
             if not repo.token:
                 return
-            try:
-                match_object = re.search('([^/]+)/([^/]+)/([^/.]+(.git)?)', repo.base)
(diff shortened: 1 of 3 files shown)


runbot/models/repo.py
```diff
@@ -123,31 +123,39 @@ def _hash_exists(self, commit_hash):
...
+                            response = session.post(url, data=json.dumps(payload))
+                        else:
+                            response = session.get(url)
+                        response.raise_for_status()
+                        return response.json()
+                    except Exception as e:
+                        exception = e
+                        messages.append('Ignored github error %s %r' % (url, payload))
+                        try_count += 1
+                        time.sleep(2)
```
Xavier-Do:
``` python3
try_count += 1
if try_count < nb_tries:
    time.sleep(2) 
else:
    if ignore_errors:
        ...
     else:
        raise e


```

Xavier-Do:
LGTM

d-fence:
@Xavier-Do change done as requested and a test written. I think this one is ready

PR #139, opened 2019-06-15
[IMP] runbot: allow to configure workers per host

d-fence:
When a runbot instance is scheduling builds, the numbers of builds
depends of a global ir.config_parameter. Even if one of the runbot
instance is running on a more powerful systsem, its number of workers is
limited by this global parameter.

With this commit, this parameter still exists but can be overriden by
specific ir.config_parameter.

For example, if the host 'runbot24.odoo.com' has more cpu power, the
number of workers for this host can be specified in the
ir.config_parameter named 'runbot24.odoo.com.workers'.

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -347,9 +347,10 @@ class runbot_repo(models.Model):
         if not ids:
             return
         icp = self.env['ir.config_parameter']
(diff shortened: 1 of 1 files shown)


Xavier-Do:
LGTM waiting for worker infos model

PR #140, opened 2019-06-17
[IMP] runbot: various improvements

Xavier-Do:


runbot/models/repo.py
```diff
@@ -366,35 +366,48 @@ def _scheduler(self):
...
+                    self.env.cr.execute(query, {'repo_ids': tuple(ids), 'host': fqdn(), 'limit': limit})
+                    return self.env.cr.fetchall()
+
+                allocated = allocate_builds("""AND runbot_build.build_type != 'scheduled'""", assignable_slots)
+                _logger.debug('Normal builds %s where allocated to runbot' % allocated)
+                weak_slot = assignable_slots - len(allocated) - 1
+                if weak_slot > 0:
+                    allocated = allocate_builds('', weak_slot)
+                    _logger.debug('Scheduled builds %s where allocated to runbot' % allocated)
+
```
d-fence: As discussed, it looks good but should be tested with ruodoo before deploying

runbot/templates/branch.xml
```diff
@@ -35,16 +33,14 @@
...
+                                <td><t t-esc="build.create_date" /></td>
                                 <td><a t-attf-href="/runbot/build/{{build['id']}}" title="Build details" aria-label="Build details"><t t-esc="build.dest" /></a></td>
                                 <td><t t-esc="build.subject" /></td>
-                                <td><t t-esc="build.name" /></td>
                                 <td><t t-esc="build.global_result" /></td>
                                 <td><t t-esc="build.global_state" /></td>
                                 <td><t t-esc="build.real_build.host" /></td>
-                                <td><t t-esc="build.active_step.name" /></td>
                                 <td><t t-esc="build.job_time" /></td>
+                                <td><t t-esc="build.build_time" /></td>
```
d-fence: Is job_time really necessary ? I think that it could add confusion with build_time

runbot/models/build.py
```diff
@@ -382,21 +382,36 @@ def _force(self, message=None):
...
+                        'extra_params': build.extra_params,
+                        'dependency_ids': build.dependency_ids,
+                        'server_match': build.server_match,
+                        'orphan_result': build.orphan_result,
+                    })
+                    #if replace: ?
+                    if build.parent_id:
+                        values.update({
+                            'parent_id': build.parent_id.id,  # attach it to parent
+                            'hidden': build.hidden,
```
d-fence: The hidden state should be copied even if the build is not a child of another. We want an **exact** rebuild no ?

runbot/templates/build.xml
```diff
@@ -67,7 +67,8 @@
                         <li><a t-attf-href="http://{{bu['domain']}}/">Connect <i class="fa fa-sign-in"></i></a></li>
                     </t>
                     <li t-if="bu.global_state in ['done','running','deathrow'] and bu_index==0" groups="base.group_user">
-                        <a href="#" class="runbot-rebuild" t-att-data-runbot-build="bu['id']">Rebuild <i class="fa fa-refresh"/></a>
+                        <a href="#" class="runbot-rebuild" t-att-data-runbot-build="bu['id']">Default Rebuild <i class="fa fa-refresh"/></a>
+                        <a href="#" class="runbot-rebuild-exact" t-att-data-runbot-build="bu['id']">Exact Rebuild <i class="fa fa-refresh"/></a>
```
d-fence: Maybe add a title to explain difference:
"Rebuild with default config and recompute branch matching" and "Rebuild with same config and same branch matching"

PR #146, opened 2019-06-27
[IMP] runbot: more human time display

KangOl:
Wait 2 unit before change.

42s ... 60s ... 119s 2m ... 60m ... 119m 2h ... 24h ... 47h 2d ...

### DIFF
--- a/runbot/common.py
+++ b/runbot/common.py
@@ -11,6 +11,9 @@ import socket
 import time
 
 from collections import OrderedDict
+from datetime import timedelta
+
+from babel.dates import format_timedelta
 
 from odoo.tools.misc import DEFAULT_SERVER_DATETIME_FORMAT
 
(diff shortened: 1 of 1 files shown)


Xavier-Do:
Approve the idea but I think we could always show two units instead. Some build are already arround two hours. What fo you think?

KangOl:
I thought about it but the display would be weird. `1h 4m` doesn't look right and is difficult to read.

We can maybe tolerate a little more that 2 units: `delay * 2.1`

Xavier-Do:
Or *10 to be consistent, and remove the unit for "decimals"
10h
9h01
1h01
59m
10m
9m59
1m01
59s
2s(59?)

xmo-odoo:
Why not use `babel.dates.format_timedelta`? It already supports rollover thresholds & minimal granularities e.g. `format_timedelta(xxx, granularity='minute', threshold=2.1)`.

KangOl:
format_timedelta is a great alternative. We can even keep the same format with `format='narrow'`

KangOl:
Can it be merged and deployed?

PR #150, opened 2019-07-08
[IMP] runbot: create a temporary symlink in Docker container

d-fence:
When odoo saves files in '/tmp' during a build, those files are lost
when the container is stopped. Typically screenshots during the HttpCase
tests.

With this commit, a directory 'tests' is created in the build dir and a
symlink is also created in /tmp, pointing to it.

With this mechanism, the files saved in '/tmp/odoo_tests' automatically
goes in buildir and survives the container removal.

### DIFF
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -38,6 +38,8 @@ def build_odoo_cmd(odoo_cmd):
     """
     # build cmd
     cmd_chain = []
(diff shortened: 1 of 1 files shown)


d-fence:
Should go alongside with this one: https://github.com/odoo/odoo/pull/34640

d-fence:
Not needed anymore

PR #152, opened 2019-07-12
[FIX] runbot: avoid traceback during run_odoo_install steps

d-fence:
If build.modules is falsy, a traceback occurs when trying to split a
boolean.

While at it, fix empty debug message during _local_cleanup and mock
_local_cleanup in test to avoid messing up with local dirs and
databases.


d-fence:
fixed in #151

PR #156, opened 2019-07-17
[FIX] runbot: clean running dockers with done builds

d-fence:
In different situations, a docker container may stay alive even if the
build global_state is done. This can lead to a build failure when a
build wants to go in running state and tries to expose the same ports as
the left over build.

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -18,7 +18,7 @@ from odoo.tools.misc import DEFAULT_SERVER_DATETIME_FORMAT
 from odoo import models, fields, api, registry
 from odoo.modules.module import get_module_resource
 from odoo.tools import config
-from ..common import fqdn, dt2time, Commit
+from ..common import fqdn, dt2time, Commit, dest_reg
 from psycopg2.extensions import TransactionRollbackError
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


runbot/models/repo.py
```diff
@@ -507,6 +507,15 @@ def _cron_fetch_and_build(self, hostname):
         """
         if hostname != fqdn():
             return 'Not for me'
+
+        # docker cleanup
+        containers = {int(dc.split('-')[0]):dc for dc in docker_ps() if '-' in dc}
+        if containers:
+            candidates = env['runbot.build'].search([('id', 'in', list(containers.keys())), ('global_state', '=', 'done')])
```
Xavier-Do: local_state?

runbot/models/repo.py
```diff
@@ -507,6 +507,15 @@ def _cron_fetch_and_build(self, hostname):
         """
         if hostname != fqdn():
             return 'Not for me'
+
+        # docker cleanup
+        containers = {int(dc.split('-')[0]):dc for dc in docker_ps() if '-' in dc}
```
Xavier-Do: split('-', 1)

Xavier-Do: could crash if docker name looks like str-str or kill docker dat are in format int-str not linked to a build.
-> use dest regex

Xavier-Do:
This need a rebase + local_state instead of global_state

PR #157, opened 2019-07-18
Master runbot classify errors moc

d-fence:
Try to classify errors seen on the runbot


runbot/models/build.py
```diff
@@ -11,6 +11,7 @@
 from ..common import dt2time, fqdn, now, grep, uniq_list, local_pgadmin_cursor, s2human, Commit
 from ..container import docker_build, docker_stop, docker_is_running, Command
 from odoo.addons.runbot.models.repo import HashMissingException
+from odoo.addons.runbot.models.build_error import clean, digest
```
Xavier-Do: no import, set methods as api.model on builderror

runbot/models/build.py
```diff
@@ -979,6 +986,42 @@ def _get_py_version(self):
...
+        BuildError = self.env['runbot.build.error']
+
+        query = """
+        SELECT build_id, name, message
+            FROM ir_logging
+            WHERE
+                type='server'
+                AND level='ERROR'
+                AND length(message) > 50
+                AND build_id NOT IN (SELECT DISTINCT build_id FROM runbot_build_error_ids_runbot_build_rel)
```
Xavier-Do: SELECT DISTINCT build_id FROM runbot_build_error_ids_runbot_build_rel WHERE build_id IN %s

Xavier-Do: Instead why not filterself first:
-> only if local_result == error
-> only if build.builderrors = False
build_to_scan = self.search(...)

xmo-odoo: Given there's probably lots of builds and errors (?) it might be more efficient to use a left outer join (and only keep the records for which there's no joined record).

runbot/models/build.py
```diff
@@ -979,6 +986,42 @@ def _get_py_version(self):
...
+                AND build_id IN %s
+        """
+        self.env.cr.execute(query, (tuple(self.ids), ))
```
xmo-odoo: FWIW one failing of "IN" is it breaks completely if given an empty tuple. =ANY(%s) and a list works fine for empty record sets as well.

runbot/models/build_error.py
```diff
@@ -0,0 +1,69 @@
...
+    responsible = fields.Many2one('res.users', 'Assigned fixer')
+    fixing_commit = fields.Char('Fixing commit')
+    build_ids = fields.Many2many('runbot.build', 'runbot_build_error_ids_runbot_build_rel',
+                                 column1='build_error_id', column2='build_id',
+                                 string='Affected builds')
+    active = fields.Boolean('Error is not fixed', default=True)
+    tag_ids = fields.Many2many('runbot.build.error.tag', string='Tags')
+    build_count = fields.Integer(compute='_compute_build_counts', string='Nb seen', stored=True)
+
+    @api.multi
```
xmo-odoo: create is @model (takes no record as input).

runbot/models/build.py
```diff
@@ -979,6 +986,42 @@ def _get_py_version(self):
...
+        ir_logs = self.env.cr.fetchall()
+        _logger.info('Logs found: %s', len(ir_logs))
+
+        hash_dict = defaultdict(list)
```
Xavier-Do: from here, should be in builderror

runbot/models/build_error.py
```diff
@@ -0,0 +1,69 @@
...
+    def create(self, vals):
+        content = vals.get('content')
+        cleaned_content = clean(content)
+        vals.update({'cleaned_content': cleaned_content})
+        vals.update({'hash': digest(cleaned_content)})
```
xmo-odoo: Either use `[]` twice, or just .update just once (also no need for the dict literal though that makes little difference).

runbot/models/build_error.py
```diff
@@ -0,0 +1,69 @@
...
+import logging
+import re
+
+from odoo import models, fields, api
+
+_logger = logging.getLogger(__name__)
+
+CLEANING_REGS = (
+    re.compile(r', line \d+,'),  # simply remove line numbers
+)
```
xmo-odoo: Using a list literal is safer here (the comma which makes a tuple into a tuple is easy to miss, and Python 1-uples are a bit ugly).

runbot/models/build_error.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def clean(s):
+    """
+    Clean the string s with the cleaning regs
+    Replacing the regex with a space
+    """
+    for r in CLEANING_REGS:
+        s = r.sub(' ', s, re.MULTILINE)
```
xmo-odoo: `Pattern.sub` doesn't take a flag, that should be passed to `re.compile`. This passes re.MULTILINE=8 as the `count` parameter (so at best it's going to stop after 8 replacements, at worst it might blow up in 3.6).

runbot/models/build_error.py
```diff
@@ -0,0 +1,92 @@
...
+    _name = "runbot.build.error"
+
+    content = fields.Text('Error message', required=True)
+    cleaned_content = fields.Text('Cleaned error message')
+    module_name = fields.Char('Module name')  # name in ir_logging
+    hash = fields.Char('Error fingerprint', index=True)
+    random = fields.Boolean('underterministic error')
+    responsible = fields.Many2one('res.users', 'Assigned fixer')
+    fixing_commit = fields.Char('Fixing commit')
+    build_ids = fields.Many2many('runbot.build', 'runbot_build_error_ids_runbot_build_rel', string='Affected builds')
```
Xavier-Do: we could add a branch_ids and repo_ids (related m2m on build_ids).
So we can easily display impacted branches/repos

runbot/templates/build.xml
```diff
@@ -99,7 +99,7 @@
                     <li><a t-attf-href="https://{{repo.base}}/compare/{{br['branch'].branch_name}}">Compare <i class="fa fa-github"/></a></li>
                     <!-- TODO branch.pull from -->
                     <li class="divider"></li>
-                    <li class="disabled"><a href="#">Port: <t t-esc="bu.real_build.port"/></a></li>
+                    <li groups="runbot.group_runbot_admin"><a t-attf-href="/web/#id={{bu['id']}}&amp;view_type=form&amp;model=runbot.build" target="new">View in backend</a></li>
```
Xavier-Do: :+1:

runbot/models/build_error.py
```diff
@@ -0,0 +1,92 @@
...
+    active = fields.Boolean('Error is not fixed', default=True)
+    tag_ids = fields.Many2many('runbot.build.error.tag', string='Tags')
+    build_count = fields.Integer(compute='_compute_build_counts', string='Nb seen', stored=True)
+
```
Xavier-Do: Maybe add a parent build_error
-> if a message is multi-line
-> if a message is different in another version

d-fence:
merged here: 0da30a9f6048

PR #162, opened 2019-07-25
[FIX] runbot: correctly order repo by sequence

xavieralt:
Even if present and editable the user, changing the sequence
of a repository has no effect

Fix by adding missing _order on runbot.repo model

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -26,6 +26,7 @@ class HashMissingException(Exception):
 class runbot_repo(models.Model):
 
     _name = "runbot.repo"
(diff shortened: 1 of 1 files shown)


xavieralt:
@d-fence quick usability fix, could you have a look ? Thanks (cc @sts-odoo)

xavieralt:
@Xavier-Do Thanks!

Xavier-Do:
I just hope it wont break anything since some cron are based on repo update order. For the main ones, the sequence order is the same as ids so I don't think it will be a problem.

PR #163, opened 2019-07-25
[IMP] runbot_merge: p=1 > split

xmo-odoo:
Allows merging a fix for e.g. a common false positive during a split
but without cancelling a staging which might just pass (you never
know).


xmo-odoo:
ping @Xavier-Do

PR #172, opened 2019-08-13
[FIX] runbot: typo (closet -> closest)

mreficent:
It seems to me a typo.


mreficent:
ping: @Xavier-Do

Xavier-Do:
It's wonderful that I managed to messed up the only point that @d-fence asked me to change.

Issue #168, opened 2019-08-06
Strip lines / ignore leading spaces when looking for commands


xmo-odoo:
https://github.com/odoo/enterprise/pull/4994#issuecomment-517691394 has a space before the `@`, and thus didn't match as a command.

Issue #167, opened 2019-07-30
Robodoo command shoud not be case sensitive


Xavier-Do:
Since username are not case sensitive making roboodo command case incensitive make sence, especially when writing Robodoo r+ from a smartphone. 

@xmo-odoo

PR #180, opened 2019-08-23
Simonis bot

xmo-odoo:
* still need to decide what user we'll be using. code may need to be modified if we decide that the fp user is the same as the mergebot's: FP currently uses "r+" as command, which would conflict. Also might need to create a new token as as FP uses somewhat different rights than mergebot.
* there's a bunch of todos & fixmes left in the code, they should be of low consequences

### DIFF
--- a/forwardport/models/project.py
+++ b/forwardport/models/project.py
@@ -1,0 +1,689 @@
+# -*- coding: utf-8 -*-
+"""
+Technically could be independent from mergebot but would require a lot of
+duplicate work e.g. keeping track of statuses (including on commits which
+might not be in PRs yet), handling co-dependent PRs, ...
+
+However extending the mergebot also leads to messiness: fpbot should have
+its own user / feedback / API keys, mergebot and fpbot both have branch
+ordering but for mergebot it's completely cosmetics, being slaved to mergebot
+means PR creation is trickier (as mergebot assumes opened event will always
+lead to PR creation but fpbot wants to attach meaning to the PR when setting
+it up), ...
+"""
+import base64
+import contextlib
+import itertools
+import json
+import logging
+import os
+import pathlib
+import subprocess
+import tempfile
+
+import re
+import requests
+
+from odoo.exceptions import UserError
+from odoo.tools.appdirs import user_cache_dir
+
+from odoo import _, models, fields, api
+from odoo.addons.runbot_merge import utils
+from odoo.addons.runbot_merge.models.pull_requests import RPLUS
+
+
+_logger = logging.getLogger('odoo.addons.forwardport')
+
+class Project(models.Model):
+    _inherit = 'runbot_merge.project'
+
+    fp_github_token = fields.Char()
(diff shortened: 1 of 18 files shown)


xmo-odoo:
@mart-e @KangOl maybe check 5d62ff3e7a9e76b2754a4272158bf6b7e8f4a15d as it changes the details of editing / rewriting PR commit messages

nim-odoo:
@KangOl

nim-odoo:
To be honest, I have no knowledge on how the runbot works at all, so I won't be able to review this in a timely manner. I'll rather let @KangOl give his feedback.

From my perspective, I'm happy as long as it does what it should :smile:

forwardport/models/project.py
```diff
@@ -0,0 +1,682 @@
...
+    fp_github_name = fields.Char(store=True, compute="_compute_git_identity")
+    fp_github_email = fields.Char(store=True, compute="_compute_git_identity")
+
+    def _find_commands(self, comment):
+        if self.env.context.get('without_forward_port'):
+            return super()._find_commands(comment)
+
+        return re.findall(
+            '^[@|#]?{}:? (.*)$'.format(self.fp_github_name),
+            comment, re.MULTILINE
```
xmo-odoo: Handle #167 (CI botname) and #168 (allow / ignore leading whitespace)

nim-odoo:
@KangOl We'd like to launch this live by the end of next week. It would be good if you have some time to have a look into it by then. Thanks!

KangOl:
For the name, I propose `abotensis` or `abotuensis`, a portmanteau between [amanuensis](https://en.wikipedia.org/wiki/Amanuensis) and bot (amanuensis is already taken on github)
In the same concept, `moine-copiste` or [`otis-odoo`](https://www.imdb.com/title/tt0250223/characters/nm0046347) are acceptable.

KangOl:
Note that there are some X-port that needs to be made when modules moves from community to enterprise (and vice-versa)
Examples:
 - `hr_payroll` moves from community to enterprise in saas-12.3: odoo/enterprise@023ec143904142d15b5d9424db1c03b044d4cee1
 - `account_asset` moves from community to enterprise in 12.0: odoo/enterprise@96546bf4cb9cfe83578c7c847e701defd151fbd4
- same for `l10n_be_intrastat`: odoo/enterprise@ba7680d93c5d236109d9ebf032d22e4a95959882
 - `sale_coupon` moves from enterprise to community in saas-12.3: odoo/odoo@60c438dfb79a847057fd51b4b05f387c70df7883

nim-odoo:
> For the name, I propose `abotensis` or `abotuensis`, a portmanteau between [amanuensis](https://en.wikipedia.org/wiki/Amanuensis) and bot (amanuensis is already taken on github)
> In the same concept, `moine-copiste` or [`otis-odoo`](https://www.imdb.com/title/tt0250223/characters/nm0046347) are acceptable.

Ouch, that is not easy to remember :confused: I'd go for `otis-odoo` which seems to be the easiest to use :smile:

PR #184, opened 2019-08-30
[FIX] runbot: limit number of allowed connections per db

d-fence:
When a runbot builds fails in a loop and creates connections in each
loop iteration, it can exhaust postgres connections.

With this commit, a build database is limited to 25 psql connections at
the same time, which should be enough.

Unfortunately, it implies that the runbot user have enough privileges to
update the datconnlimit column in pg_database table.


Xavier-Do:
need specific privilege right, canceled for now

PR #188, opened 2019-09-05
[IMP] runbot: various improvements

d-fence:
- Add a keep running flag on the build to allow a build to stay in
  running state until the flag is switched off ( or the build killed)
- Do not update configs and config_steps data
- Add a first/last_seen_build and first/last_seen_date on build.error
- Children error builds now include the parent builds too
- Use a notebook on build.error form view to display builds and linked
  errors
- Update result when a build triggers a change from 'warn' to 'ko' too

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -107,6 +107,7 @@ class runbot_build(models.Model):
                                         string='Source export path mode')
     build_url = fields.Char('Build url', compute='_compute_build_url', store=False)
     build_error_ids = fields.Many2many('runbot.build.error', 'runbot_build_error_ids_runbot_build_rel', string='Errors')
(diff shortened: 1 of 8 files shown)


runbot/models/build.py
```diff
@@ -607,7 +608,7 @@ def _schedule(self):
             else:  # testing/running build
                 if build.local_state == 'testing':
                     # failfast in case of docker error (triggered in database)
-                    if (not build.local_result or build.local_result == 'ok') and build.triggered_result:
+                    if (not build.local_result or build.local_result in ['ok', 'warn']) and build.triggered_result:
```
Xavier-Do: do not notify github in warn->warn case

mreficent:
remove repeated call in `runbot/tests/__init__.py` 😉

PR #189, opened 2019-09-08
[IMP] runbot: number of log lines per build

d-fence:
When a build is running, a cron, an evil query or something else can
start to fill and bloat the runbot ir_logging table.

With this commit, a log_counter field is added on the build, starting at
100. The SQL trigger decrement this counter after a line is inserted.

When the counter drops to 0, a the last log line contains a message
stating that the limit has been reached. Further log lines are dropped
for this build step.

The counter is reset to 100 before each step.

The runbot itself is still able to add logs lines through the build _log
method.

Thanks @Xavier-Do for the smart idea.

### DIFF
--- a/runbot/tests/test_event.py
+++ b/runbot/tests/test_event.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+from unittest.mock import patch
+from odoo.tests import common
+
+
+class TestIrLogging(common.TransactionCase):
+
+    def setUp(self):
+        super(TestIrLogging, self).setUp()
+        self.Repo = self.env['runbot.repo']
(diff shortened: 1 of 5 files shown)


runbot/tests/test_event.py
```diff
@@ -0,0 +1,68 @@
...
+        build.invalidate_cache()
+        self.assertEqual(build.triggered_result, 'warn', 'A warning log should sets the build in warn')
+
+        #  Test that a error log line sets the build in ko
+        self.simulate_log(build, 'test function', 'test message', level='ERROR')
+        build.invalidate_cache()
+        self.assertEqual(build.triggered_result, 'ko', 'An error log should sets the build in ko')
+        
+        # Test the log limit
+        for i in range(1000):
```
Xavier-Do: Manually set log_counter to 10 so that test will be faster.

d-fence:
@Xavier-Do rebased

PR #192, opened 2019-09-10
[FIX] forwardport: clarify forward port message

mart-e:
The message was not clear
- don't know if anything is needed from the author
- don't know if the last step in the chain

Remove ping message and include it in the first one,

Maybe a documentation page would clarify it
_edit: done https://github.com/odoo/odoo/wiki/Mergebot#forward-port_

### DIFF
--- a/forwardport/models/project.py
+++ b/forwardport/models/project.py
@@ -387,24 +387,6 @@ class PullRequests(models.Model):
             message += "Forward-Port-Of: %s#%s" % (source.repository.name, source.number)
 
             (h, out, err) = conflicts.get(pr) or (None, None, None)
-            if h:
-                message = """Cherrypicking %s of source #%d failed with the following
-
-stdout:
-```
-%s
-```
-
-stderr:
-```
-%s
-```
-
-Either perform the forward-port manually (and push to this branch, proceeding
-as usual) or close this PR (maybe?).
-
-In the former case, you may want to edit this PR message as well.
-""" % (h, source.number, out, err)
 
             r = requests.post(
                 'https://api.github.com/repos/{}/pulls'.format(pr.repository.name), json={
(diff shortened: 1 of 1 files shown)


xmo-odoo:
The issue I have with this is the PR message becomes a commit in case of merge or rebase-merge, so the more it strays from the original the more garbage we will end up with.

forwardport/models/project.py
```diff
@@ -433,11 +415,58 @@ def _port_forward(self):
...
+
+To merge the full chain, say
+> @fw-bot r+
+
+More info at
+https://github.com/odoo/odoo/wiki/Mergebot#forward-port
+""" % (target.name, chain_msg)
+            else:
+                message = """
+This PR targets %s and is part of the forard-port chain.
```
KangOl:
```suggestion
This PR targets %s and is part of the forward-port chain.
```

forwardport/models/project.py
```diff
@@ -433,11 +415,58 @@ def _port_forward(self):
...
+Either perform the forward-port manually (and push to this branch, proceeding
+as usual) or close this PR (maybe?).
+
+In the former case, you may want to edit this PR message as well.
+""" % (h, source.number, out, err)
+            elif base.limit_id == target:
+                chain, current = [pr.number], pr
+                while current.parent_id and current.parent_id != source:
+                    chain.append(current.parent_id)
+                    current = current.parent_id
```
xmo-odoo: this could probably use _iter_ancestors

forwardport/models/project.py
```diff
@@ -433,11 +415,58 @@ def _port_forward(self):
...
+                chain_msg = "\n* ".join("#%s" % c.number for c in chain)
+                message = ping + """
+This PR targets %s and is the last of the forward-port chain containing:
+* %a
```
xmo-odoo: why %a?

mart-e: because typo

mart-e:
Fixed but obviously not tested

xmo-odoo:
> Fixed but obviously not tested

400px-GLaDOS_P2

PR #199, opened 2019-09-14
[IMP] runbot: create a new build error when a fixed one reappear

d-fence:
When a build error appears with the same fingerprint as already known
one which was supposedly fixed, the build is simply added to the known
build error.

In order to keep an eye on such reappearing bugs and keep the fixing
history separated, this commit simply creates a new build_error and link
the old one to it.


runbot/models/build_error.py
```diff
@@ -91,19 +91,22 @@ def _parse_logs(self, ir_logs):
...
         for fingerprint, logs in hash_dict.items():
-            self.env['runbot.build.error'].create({
+            build_error = self.env['runbot.build.error'].create({
                 'content': logs[0].message,
                 'module_name': logs[0].name,
                 'function': logs[0].func,
                 'build_ids': [(6, False, [r.build_id.id for r in logs])],
             })
+            if build_error.fingerprint in fixed_errors_dict:
+                fixed_errors_dict[build_error.fingerprint].write({'parent_id': build_error.id})
```
d-fence: we should take care of build_errors that already have a parent_id set.
Maybe we should link the parent instead.

Xavier-Do: I think it should be another field, maybe just a computed field to now old errors with same hash

```
_compute_error_history(self):
    for error in self:
        error_history = self.search([('fingerprint', '=', error.fingerprint)])
```

runbot/tests/__init__.py
```diff
@@ -1,7 +1,9 @@
 from . import test_repo
+from . import test_build_error
 from . import test_branch
 from . import test_build
 from . import test_frontend
 from . import test_schedule
 from . import test_cron
 from . import test_build_config_step
+from . import test_build_error
```
mreficent: repeated

PR #208, opened 2019-09-23
[IMP] runbot: uses fnmatch to black list modules

d-fence:
Actually some Odoo modules are black_listed from a set hardcoded in the
runbot code. On some cases, one needs to blacklist custom modules,
preferably in a config_step.

With this commit, a mod_black_list field is added on the Repo and on the
ConfigStep. They optionaly hold a comma separated list of fnmatch patterns. If a
module name match any of the pattern, it's filtered out from the build.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import fnmatch
 import glob
 import logging
 import os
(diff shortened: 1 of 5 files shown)


runbot/models/build.py
```diff
@@ -693,21 +693,21 @@ def _server(self, *path):
...
-        blacklist_modules = set(['auth_ldap', 'document_ftp', 'base_gengo',
-                                 'website_gengo', 'website_instantclick',
-                                 'pad', 'pad_project', 'note_pad',
-                                 'pos_cache', 'pos_blackbox_be'])
+    def _filter_modules(self, modules, available_modules, explicit_modules, exclude_regexes):
+        """Filters modules
+        :param modules: list of modules candidates
+        :param available_modules: list of availables modules
+        :param explicit modules: list of modules explicitely defined
+        :param exclude_regexes: comma separated list of regexes to exlcude modules
```
mreficent:
```suggestion
        :param exclude_regexes: comma separated list of regexes to exclude modules
```

runbot/models/repo.py
```diff
@@ -122,6 +123,18 @@ def _get_repo_name_part(self):
...
+    def _check_mod_black_list(self):
+        for repo in self:
+            for pat in repo.mod_black_list.split(','):
+                error = None
+                try:
+                    re.compile(pat)
+                except re.error as e:
+                    error = e
+                if error:
+                    raise ValidationError(error)
```
xmo-odoo: Why not just directly `raise ValidationError(error)` from the exception handler, and remove the local variable?

PyXD: I think it is to avoid to have two stack trace but "raise ValidationError(e)" from None should do the trick.

Xavier-Do:
Talking to @JKE-be, we also have the case of design theme where we want to execute all website_*
So we may want to consider using  regex for module list too.

Optionnal: I think it would maybe be better to use fnmatch instead of regex since we don't really need regex features, and fnmatch syntax would be cleaner. 
"website_*" instead of "website_.*"

This would also help if we want to add everything in a single field prefixing black_list with minus.

The spec for this would be :
design theme: 
-modules: website_*, test_website
-module_auto: all (instead of none)

odoo odoo:
-modules: *, -hw_*, -theme_*, -l10n_*, -auth_ldap, -document_ftp, -base_gengo, -website_gengo, -website_instantclick, -pad, -pad_project, -note_pad, -pos_cache, -pos_blackbox_be
-module_auto: all

Code would be pretty simple I think since all we need is to find module list based on module_auto (repo or all) and filter them usinf fnmatch. Keeping modules empty or with explicit list will correspond to modules_auto none, default value for modules would be "*".

Issue #218, opened 2019-09-30
If a forward-port PR gets staged and that staging gets cancelled (/failed?) it gets wedged


xmo-odoo:
Issue occurred on odoo/enterprise#5822 and it locked up merging stagings as it would blow up validation.

Also todo: fix up validation issues.

Issue #213, opened 2019-09-27
Handling of "up to" on FP PRs and / or after the FP process has started


xmo-odoo:
Currently possible to `@fw-bot up to x` on one of the forward-port PRs, but fw-bot acknowledges it then does nothing. Also not sure the command will be taken in account if the forward-porting process has already started, even if set on the source PR.

See odoo/enterprise#5781

Issue #212, opened 2019-09-27
Incomplete message on final ping


xmo-odoo:
odoo/odoo#37452 seems to have been forward-ported from 12.0 to 12.5 through 12.3 and 12.4 without any human intervention, however it only lists the previous PR in the chain (37447) but not the one before that (37440).

Also there's a typo in the message (missing space between "chain" and "containing").

Issue #211, opened 2019-09-27
Prevent moving stagings between success and failure


xmo-odoo:
If a staging is rebuilt, or a staging was explicitly rejected or approved, if the runbot build is not cancelled it will eventually send a status (usually error / failure as staging branches are not sticky) which is going to update the staging's state, which can have odd side-effects (e.g. a PR being flagged as error after being merged).

The success / failure state of a staging should be terminal.

PR #219, opened 2019-10-01
[FIX] forwardport: add missing space

mart-e:
cf odoo/odoo#37628

> This PR targets saas-12.5 and is the last of the forward-port chaincontaining:


xmo-odoo:
@mart-e I have a local fix with a few other issues (cf #212, I'm also fixing #210 and #211, and currently working on #213). However these are relatively minor and cosmetic issues, so I figured I'd merge them after the big rush. 

I need to fix #218 first though. I've already pushed a workaround-ish (in that a wedged PR won't break the entire cron job) but I need to fix the underlying issue so we don't get FP PRs in that state.

PR #240, opened 2019-10-24
[IMP] runbot: check git repos on hosts

d-fence:
From time to time, a git repository is found corrupted on a runbot.
In that case, it can lead to various unpredictable resulsts.

For example, git archive fails or a commit is unreachable, leading to a
runbot build with no Odoo server. Even worse, the git archive can lead
to a partial extract of the tree with some files missing and to wrong
failure interpretation.

With this commit, at the end of the fetch_and_build cron, each active
repo is searched on the host to find a "gc.log" file.
If found, the last_exception field is used to store the content of the
gc.log file and the host is reserved, that way the host will not starts
new builds until manually fixed.

Note:
As a further improvement, a repo._reclone method could be executed to
reclone the repo from scratch in such a case and unreserve the host at
the end of the process.

### DIFF
--- a/runbot/models/host.py
+++ b/runbot/models/host.py
@@ -48,6 +48,16 @@ class RunboHost(models.Model):
         name = fqdn()
         return self.search([('name', '=', name)]) or self.create({'name': name})
 
+    def _check_repos(self):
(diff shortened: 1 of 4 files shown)


d-fence:
done in another PR

PR #243, opened 2019-11-05
[FIX] runbot: fix container cli tests

d-fence:
Since 81fefee, the container.py CLI does not work as expected.

With this commit, the CLI is working, a new arg was added to test
flamegraphs and the dump is adapted to mimic the runbot.

Also, a small issue is fixed in the zip file creation. Before the zip
creation, the directory is changed, if the directory change fails, the
zip is created from the current directory which is removed by zip at the
end of the process. That could lead to the deletion of the build dir.


d-fence:
@Xavier-Do I think it's ready

PR #251, opened 2019-11-08
[IMP] runbot: use a config file to simplify args

d-fence:
When starting an odoo instance with Docker, a very long command line is
computed and appears in the logs.

With this commit, an .odoorc configuration file is written ind the build
dir and mounted in the Docker container.

Previously, the runbot .odoorc/.openerprc file was mounted to share some
parameters. Now, if that file exsists, its content is merged with build .odoorc.

### DIFF
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -9,7 +9,9 @@ When testing this file:
     The second parameter is the exposed port
 """
 import argparse
+import configparser
 import datetime
+import io
 import json
 import logging
 import os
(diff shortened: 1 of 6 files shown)


runbot/container.py
```diff
@@ -32,11 +34,20 @@
...
+        :param posts: list of post commands posts only run if the cmd command succedd (&&)
+        :param finals: list of finals commands always executed
+        :param rc: list of key,value tuples to write in config file
+        returns a string of the full command line to run
+        """
         self.pres = pres or []
         self.cmd = cmd
         self.posts = posts or []
         self.finals = finals or []
+        self.rc = rc or []
```
Xavier-Do: I would prefer a self.configs and def add_config, because I don't think that a Command needs remotes controls or Rubik cube.

PR #252, opened 2019-11-10
[IMP] runbot: allow to change prefix in multi wizard

d-fence:


d-fence:
@Xavier-Do very small improvement

PR #253, opened 2019-11-11
[IMP] runbot: refactor tests

d-fence:
A lot of things have to be mocked during runbot tests, as a consequence,
a lot of patch decorators accumulate in a big stack uppon some tests
methods.

Also, a lot of mocks are used multiple times among tests.

With this commit, a new RunbotClass is added that comes with patches
ready to be started. A start_patcher helper method is available to start
a patch and add the appropriate stop in a cleanup.

Also, when a build is created in the tests, the _get_params method is
always called, resulting in an annoying git warning.

With this commit, a create_build method is added on the test class, that
way the _get_params is always mocked when a build is created.

### DIFF
--- a/runbot/tests/common.py
+++ b/runbot/tests/common.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+from odoo.tests.common import TransactionCase
+from unittest.mock import patch
+
+class Dummy():
+    ...
+
+
+class RunbotCase(TransactionCase):
+
+    def setUp(self):
+        super(RunbotCase, self).setUp()
+
+        self.Build = self.env['runbot.build']
+        self.Repo = self.env['runbot.repo']
+        self.Branch = self.env['runbot.branch']
+
+        self.patchers = {}
+
+        def git_side_effect(cmd):
+            if cmd[:2] == ['show', '-s'] or cmd[:3] == ['show', '--pretty="%H -- %s"', '-s']:
+                return 'commit message for %s' % cmd[-1]
+            if cmd[:2] == ['cat-file', '-e']:
+                return True
+            else:
+                print('Unsupported mock command %s' % cmd)
+
+        self.start_patcher('git_patcher', 'odoo.addons.runbot.models.repo.runbot_repo._git', side_effect=git_side_effect)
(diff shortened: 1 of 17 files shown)


runbot/tests/common.py
```diff
@@ -12,27 +15,45 @@ def setUp(self):
...
-        mock_get_params.return_value = defaultdict(lambda: defaultdict(str))
-        return mock_get_params
+            self.patchers[patcher_name] = res
+            if side_effect != Dummy:
+                res.side_effect = side_effect
+            elif return_value != Dummy:
+                res.return_value = return_value
+            else:
+                def _side_effect(*args, **kwargs):
+                    print(patcher_name, 'calledwith', args, kwargs)
```
d-fence: debug print ?

runbot/tests/test_cron.py
```diff
@@ -25,46 +24,40 @@ def test_crons_returns(self):
...
     @patch('odoo.addons.runbot.models.repo.runbot_repo._reload_nginx')
     @patch('odoo.addons.runbot.models.repo.runbot_repo._scheduler')
-    def test_cron_build(self, mock_scheduler, mock_reload, mock_cron_period, mock_host_fqdn):
+    def test_cron_build(self, mock_scheduler, mock_reload):
         """ test that cron_fetch_and_build do its work """
-        hostname = 'runbotx.foo.com'
-        self.mock_fqdn.return_value = mock_host_fqdn.return_value = hostname
-        mock_cron_period.return_value = 2
+        hostname = 'host.runbot.com'
+        #self.mock_fqdn.return_value = mock_host_fqdn.return_value = hostname
```
d-fence: remove

runbot/tests/test_repo.py
```diff
@@ -226,8 +228,8 @@ def setUp(self  ):
     @patch('odoo.addons.runbot.models.build.runbot_build._kill')
     @patch('odoo.addons.runbot.models.build.runbot_build._schedule')
     def test_repo_scheduler(self, mock_schedule, mock_kill, mock_reap):
-        mock_fqdn = self.start_patcher('fqdn_patcher')
-        mock_fqdn.return_value = 'test_host'
+        #mock_fqdn = self.patchers['fqdn_patcher']
+        #mock_fqdn.return_value = 'test_host'
```
d-fence: remove

PR #261, opened 2019-11-29
Fix useless redirections on build actions

xmo-odoo:
Killing, waking or rebuilding currently redirects to the repository root which is generally useless.

Since it's very easy to go from a build to its repository, redirect to the "most useful" build instead:

* the current build for kill & wakeup
* the new build on rebuild

Provided two versions of the implementation:
* one keeps the current controllers & deduplicates and fixes the final URL building
* the other one merges all 3 controllers into a single one

### DIFF
--- a/runbot/controllers/frontend.py
+++ b/runbot/controllers/frontend.py
@@ -3,8 +3,12 @@ import operator
 import werkzeug
 from collections import OrderedDict
 
+import werkzeug.utils
+import werkzeug.urls
+
 from odoo.addons.http_routing.models.ir_http import slug
 from odoo.addons.website.controllers.main import QueryURL
+
 from odoo.http import Controller, request, route
 from ..common import uniq_list, flatten, fqdn
 from odoo.osv import expression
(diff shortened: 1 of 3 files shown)


xmo-odoo:
Updated version:
* rewrote JS to do xhr
* maps actions to either following the redirection or reloading the current page, for convenience killing a build now reloads the page in-place, rebuilding opens the new build. Waking also opens the new build though it could reload instead I guess.

xmo-odoo:
An alternative change would be to:
* make all the action links into proper forms / buttons (note that this is harder to style), removing the need for JS
* have the forms optionally provide a redirection destination for kill and wakeup
* have the endpoint redirect to the provided URL or the (new) build-url

xmo-odoo:
Better specs:

* if interacting with a build while not on that specific build's page, just reload whatever page we're on e.g. killing or rebuilding from the repo, or waking a sub-build from a parent
* if interacting with a build from that specific build's page, then
  - reload current page if killing or waking
  - redirect to new page if rebuild or exact rebuild

xmo-odoo:
@Xavier-Do should be fixed but I really don't really know how to run it on a "filled" system so I can check that there's no errors and all the redirects are to the right target.

PR #266, opened 2019-12-06
Hook times in another table

Xavier-Do:


runbot/models/repo.py
```diff
@@ -77,6 +77,51 @@ def _inverse_config_id(self):
...
+        self.env.cr.execute("""
+            DELETE from runbot_repo_reftime WHERE id NOT IN (
+                SELECT max(id) FROM runbot_repo_reftime GROUP BY repo_id
+            )
+        """, (tuple(self.ids),))
+        self.env.cr.execute("""
+            DELETE from runbot_repo_hooktime WHERE id NOT IN (
+                SELECT max(id) FROM runbot_repo_hooktime GROUP BY repo_id
+            )
+        """, (tuple(self.ids),))
```
xmo-odoo: ids don't seem used in query

runbot/models/repo.py
```diff
@@ -77,6 +77,51 @@ def _inverse_config_id(self):
...
+            self.env['runbot.repo.reftime'].create({'time': repo.get_ref_time, 'repo_id': repo.id})
+
+    def _compute_hook_time(self):
+        self.env.cr.execute("""
+            SELECT repo_id, time FROM runbot_repo_hooktime
+            WHERE id IN (
+                SELECT max(id) FROM runbot_repo_hooktime 
+                WHERE repo_id in %s GROUP BY repo_id
+            )
+        """, (tuple(self.ids),))
```
xmo-odoo: can use `=any(%s)` instead of `in %s` so you don't have to convert to a tuple, and as bonus it works correcly if the input is empty.

runbot/models/repo.py
```diff
@@ -77,6 +77,51 @@ def _inverse_config_id(self):
...
+        times = dict((r[0], r[1]) for r in self.env.cr.fetchall())
```
xmo-odoo: Can just call `dict(cr.fetchall())`.

runbot/models/repo.py
```diff
@@ -77,6 +77,57 @@ def _inverse_config_id(self):
...
+        # hooktime and reftime table are here to avoid sql update on repo.
+        # using inverse will still trigger write_date and write_uid update.
+        # this hack allows to avoid that
+        for repo in self:
+            if 'hook_time' in values:
+                hook_time = values.pop('hook_time')
+                self.env['runbot.repo.hooktime'].create({'time': hook_time, 'repo_id': repo.id})
+            if 'get_ref_time' in values:
+                get_ref_time = values.pop('get_ref_time')
+                self.env['runbot.repo.reftime'].create({'time': get_ref_time, 'repo_id': repo.id})
```
xmo-odoo: You need to check for & pop values outside the loop, otherwise it's going to work for the first repo and not for any other. Also could probably pop the hook_time directly (`values.pop('hook_time', None)`), and if it's truthy then do the thing, not sure there's any reason to create a hook time or ref time with no time data.

Might also be a good idea to require the time while at it.

PR #270, opened 2019-12-12
[IMP] runbot: show sub-builds logs on parent build

d-fence:
Now that we can split build in multiple children builds, it can be confusing to read/find logs.

This PR is an attempt to show the logs of the sub-builds on the parent build log page.

### DIFF
--- a/runbot/controllers/frontend.py
+++ b/runbot/controllers/frontend.py
@@ -160,28 +160,14 @@ class Runbot(Controller):
         if not build.exists():
             return request.not_found()
 
-        # other builds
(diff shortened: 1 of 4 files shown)


d-fence:
@Xavier-Do I'm not even sure that it is a good idea to do that at all.

- Logs are coming in unpredictable order, depending the time the sub-build started.
- That puts a lot of informations on a single page, forcing the user to sort that mess by himself
- access right problems on ir.logging for a public user
- filtering may be difficult to put in place again
...
does it worth the effort ?

PR #271, opened 2019-12-13
[HACK] runbot: create a broken symlink in all server sources

d-fence:
As the --upgrade-paths options does not work as expected in Odoo, a
symbolic link as to be created in odoo/addons/base/maintenance pointing
to the migration scripts.

The runbot uses Docker read-only volumes to access the sources that are
shared between builds, preventing the creation of such a link.

With this commit, a symbolic link is created right after the export of a
commit only when the repo is a "server" repo.

This link is broken outside of the Docker volumes but uses the mount
points of the sources inside the container.

Two ir.config_parameter's are used to enable and configure this feature:

* runbot_migration_ln: the relative path where the link should be placed
  (typically odoo/addons/base/maintenance)

* runbot_migration_repo_id: the id of the migration scripts repo, used
to determine the name of the mount point inside the Docker container

### DIFF
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -212,6 +212,14 @@ class runbot_repo(models.Model):
         if err:
             raise RunbotException("Archive %s failed. Did you force push the branch since build creation? (%s)" % (sha, err))
 
+        # migration scripts link if necessary
+        icp = self.env['ir.config_parameter']
+        ln_param = icp.get_param('runbot_migration_ln', default='')
+        migration_repo_id = int(icp.get_param('runbot_migration_repo_id', default=0))
+        if ln_param and migration_repo_id and self.server_files:
+            scripts_dir = self.env['runbot.repo'].browse(migration_repo_id)._get_repo_name_part()
+            os.symlink('/data/build/%s' % scripts_dir,  self._source_path(sha, ln_param))
+
         # TODO get result and fallback on cleaing in case of problem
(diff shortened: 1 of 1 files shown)


runbot/models/repo.py
```diff
@@ -212,6 +212,14 @@ def _git_export(self, sha):
         if err:
             raise RunbotException("Archive %s failed. Did you force push the branch since build creation? (%s)" % (sha, err))
 
+        # migration scripts link if necessary
+        icp = self.env['ir.config_parameter']
+        ln_param = icp.sudo().get_param('runbot_migration_ln', default='')
+        migration_repo_id = int(icp.sudo().get_param('runbot_migration_repo_id', default=False))
```
Xavier-Do: default 0 (and no sudo)

runbot/models/repo.py
```diff
@@ -212,6 +212,14 @@ def _git_export(self, sha):
...
+        if ln_param and migration_repo_id and self.server_files:
+            scripts_dir = self.env['runbot.repo'].browse(migration_repo_id)._get_repo_name_part()
+            os.symlink('/data/build/%s' % scripts_dir, '/data/build/%s/%s' % (self._get_repo_name_part(), ln_param))
```
Xavier-Do: I think it should be something like  os.symlink('/data/build/%s' % scripts_dir, self._source_path(sha, *path))

PR #278, opened 2019-12-21
[IMP] runbot: add a data field on build

d-fence:
When two steps in the same build needs to exchange informations, some
hacks have to be used. E.g. using the extra_params fields to store comma
separated values.

With this commit, a data field is added alongside with two helper
methods that store/retrieve the content of the field as a json string.

### DIFF
--- a/runbot/fields.py
+++ b/runbot/fields.py
@@ -1,0 +1,50 @@
+from odoo.fields import Field, pycompat
+import json
+from collections import MutableMapping
+from psycopg2.extras import Json
+
+
+class JsonDictField(Field):
+    type = 'jsonb'
+    column_type = ('jsonb', 'jsonb')
+    column_cast_from = ('varchar',)
+
+    def convert_to_write(self, value, record):
+        return value
+
+    def convert_to_column(self, value, record, values=None, validate=True):
+        val = self.convert_to_cache(value, record, validate=validate)
+        return Json(value) if val else val
+
+    def convert_to_cache(self, value, record, validate=True):
+        return value.dict if isinstance(value, FieldDict) else value if isinstance(value, dict) else None
+
+    def convert_to_record(self, value, record):
+        return FieldDict(value or {}, self, record)
+
+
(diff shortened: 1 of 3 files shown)


Xavier-Do:
It doesn't need to be private methods I think but LGTM

d-fence:
@Xavier-Do I did a "property/setter" version which seems more elegant to me.
I prefer this version but it may look overkill for a so simple field.

Xavier-Do:
It's a verry good idea, we will be able to do stuff like:

```python3
build.data['restore_url'] = '...'
```
We just need a default value, I think we may enforce to have dict in data, and have a default value to '{}' So that json.loads doesn't fail when reading the field if not set. using _data or '{}' could also work in getter to avoid to write on all build on next -u. 

Some small suggestion: 
- I would personally rename _data to json_data or serialized_data
- I'm not sure if we should use repr and literal_eval or json.  Json should be fine.
- I wonder if data is not a little to generic, but it's good enough for me.

d-fence:
@Xavier-Do

> It's a very good idea, we will be able to do stuff like:
> 
> ```python
> build.data['restore_url'] = '...'
> ```

No, it won't work. I think that we may have to implement a ["dictproperty' ](http://code.activestate.com/recipes/440514-dictproperty-properties-for-dictionary-attributes/).
It's probably overkill for what we need.
But we will still be able to do:
 ```python
build.json_data = {'restore_url' : '...'}
```
 Which will store that as json.

> We just need a default value, I think we may enforce to have dict in data, and have a default value to '{}' So that json.loads doesn't fail when reading the field if not set. using _data or '{}' could also work in getter to avoid to write on all build on next -u.
> 
> Some small suggestion:
> 
>     * I would personally rename _data to json_data or serialized_data

I choose "_json_data" and "json_data" for the property. I think we have to be symmetric between the property and the field.

>     * I'm not sure if we should use repr and literal_eval or json.  Json should be fine.
> 
>     * I wonder if data is not a little to generic, but it's good enough for me.

PR #279, opened 2019-12-24
[FIX] runbot: correctly import os when using cli

d-fence:
When container.py is used from command line, it crashes because it
cannot import os from the "common" file.

The container started by the CLI now use the new docker_state function
to check if the container is running or has finished.

Also, this commit adds test-enable params to the container.py script which is
symmetric to the same Odoo CLI parameter.

### DIFF
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -14,7 +14,10 @@ import datetime
 import io
 import json
 import logging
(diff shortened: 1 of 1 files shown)


d-fence:
Already fixed in 45799df0942

PR #286, opened 2020-01-07
[IMP] runbot: improve local cleanup

d-fence:
When a build age reaches the gc_days parameter, its database is dropped
and its directory is removed.

With this commit, two fields are added in order to keep some builds
longer that the defined gc_days.

The gc_delay field on the build allows to add a delay (in number of
days) that is added to its gc_days to compute the gc_date.

The gc_date field is the date when the cleaning will occur.

Also, a test is added and the RunbotCase test class is improved to allow
the stop of a patcher.

### DIFF
--- a/runbot/common.py
+++ b/runbot/common.py
@@ -97,6 +97,7 @@ def s2human(time):
         threshold=2.1,
     )
 
+
 @contextlib.contextmanager
 def local_pgadmin_cursor():
     cnx = None
@@ -107,3 +108,17 @@ def local_pgadmin_cursor():
     finally:
         if cnx:
             cnx.close()
+
+
+def list_local_dbs(additionnal_conditions=None):
+    additionnal_condition_str = ''
+    if additionnal_conditions:
+        additionnal_condition_str = 'AND (%s)' % ' OR '.join(additionnal_conditions)
+    with local_pgadmin_cursor() as local_cr:
+        local_cr.execute("""
+            SELECT datname
+                FROM pg_database
+                WHERE pg_get_userbyid(datdba) = current_user
+                %s
+        """ % additionnal_condition_str)
+        return [d[0] for d in local_cr.fetchall()]
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -8,7 +8,7 @@ import shutil
 import subprocess
 import time
 import datetime
-from ..common import dt2time, fqdn, now, grep, uniq_list, local_pgadmin_cursor, s2human, Commit, dest_reg, os
+from ..common import dt2time, fqdn, now, grep, local_pgadmin_cursor, s2human, Commit, dest_reg, os, list_local_dbs
 from ..container import docker_build, docker_stop, docker_state, Command
 from ..fields import JsonDictField
 from odoo.addons.runbot.models.repo import RunbotException
@@ -99,6 +99,8 @@ class runbot_build(models.Model):
     job = fields.Char('Active step display name', compute='_compute_job')
     job_start = fields.Datetime('Job start')
     job_end = fields.Datetime('Job end')
+    gc_date = fields.Datetime('Local cleanup date', compute='_compute_gc_date')
+    gc_delay = fields.Integer('Cleanup Delay', help='Used to compute gc_date')
     build_start = fields.Datetime('Build start')
     build_end = fields.Datetime('Build end')
     job_time = fields.Integer(compute='_compute_job_time', string='Job time')
@@ -164,6 +166,17 @@ class runbot_build(models.Model):
                         record.global_state = 'waiting'
                 else:
                     record.global_state = record.local_state
+
+    @api.depends('gc_delay', 'job_end')
+    def _compute_gc_date(self):
+        icp = self.env['ir.config_parameter']
+        max_days_main = int(icp.get_param('runbot.db_gc_days', default=30))
+        max_days_child = int(icp.get_param('runbot.db_gc_days_child', default=15))
+        for build in self:
+            ref_date = fields.Datetime.from_string(build.job_end or build.create_date or fields.Datetime.now())
+            max_days = max_days_main if not build.parent_id else max_days_child
+            max_days += int(build.gc_delay if build.gc_delay else 0)
+            build.gc_date = ref_date + datetime.timedelta(days=(max_days))
 
     def _get_youngest_state(self, states):
         index = min([self._get_state_score(state) for state in states])
@@ -496,10 +509,6 @@ class runbot_build(models.Model):
         return self.browse()
 
     def _filter_to_clean(self, dest_list, label):
-        icp = self.env['ir.config_parameter']
-        max_days_main = int(icp.get_param('runbot.db_gc_days', default=30))
-        max_days_child = int(icp.get_param('runbot.db_gc_days_child', default=15))
-
         dest_by_builds_ids = defaultdict(list)
         ignored = set()
         for dest in dest_list:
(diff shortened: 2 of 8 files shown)


runbot/models/repo.py
```diff
@@ -469,6 +460,7 @@ def _commit(self):
     def _scheduler(self, host):
         nb_workers = host.get_nb_worker()
 
+        self._gc_testing(host)
```
Xavier-Do: add commit after that?

runbot/models/repo.py
```diff
@@ -550,6 +542,28 @@ def _gc_running(self, host):
...
+        used_slots = len(testing_builds)
+        available_slots = host.get_nb_worker() - used_slots
+        nb_pending = Build.search_count([('local_state', '=', 'pending'), ('host', '=', False)])
+        if available_slots > 0 or nb_pending == 0:
+            return
+
+        builds_to_kill = self.env['runbot.build']
+        for build in testing_builds:
+            if not build.branch_id.sticky:
+                newer_candidates = Build.search([('id', '>', build.id), ('branch_id', '=', build.branch_id.id), ('build_type', '=', 'normal'), ('parent_id', '=', False), ('local_state', '=', 'testing')])
```
Xavier-Do: -global_state not running, done, or ignore state
-not hidden
-config_id = build._get_top_parent.config_id

runbot/models/repo.py
```diff
@@ -550,6 +542,28 @@ def _gc_running(self, host):
         build_ids = Build.search(domain_host + [('local_state', '=', 'running'), ('id', 'not in', cannot_be_killed_ids)], order='job_start desc').ids
         Build.browse(build_ids)[running_max:]._kill()
 
+    def _gc_testing(self, host):
+        """garbage collect builds that could be killed"""
+        # decide if we need room
+        Build = self.env['runbot.build']
+        domain_host = self.build_domain_host(host)
+        testing_builds = Build.search(domain_host + [('local_state', '=', 'testing')])
```
Xavier-Do: testing/pending

runbot/models/repo.py
```diff
@@ -550,6 +542,28 @@ def _gc_running(self, host):
...
+        available_slots = host.get_nb_worker() - used_slots
+        nb_pending = Build.search_count([('local_state', '=', 'pending'), ('host', '=', False)])
+        if available_slots > 0 or nb_pending == 0:
+            return
+
+        builds_to_kill = self.env['runbot.build']
+        for build in testing_builds:
+            if not build.branch_id.sticky:
+                newer_candidates = Build.search([('id', '>', build.id), ('branch_id', '=', build.branch_id.id), ('build_type', '=', 'normal'), ('parent_id', '=', False), ('local_state', '=', 'testing')])
+                if newer_candidates.exists():
```
Xavier-Do: remove exists

runbot/common.py
```diff
@@ -107,3 +108,17 @@ def local_pgadmin_cursor():
     finally:
         if cnx:
             cnx.close()
+
+
+def list_local_dbs(additionnal_conditions=()):
```
Xavier-Do: =None

runbot/models/repo.py
```diff
@@ -550,6 +543,36 @@ def _gc_running(self, host):
...
+        domain_host = self.build_domain_host(host)
+        testing_builds = Build.search(domain_host + [('local_state', 'in', ['testing', 'pending'])])
+        used_slots = len(testing_builds)
+        available_slots = host.get_nb_worker() - used_slots
+        nb_pending = Build.search_count([('local_state', '=', 'pending'), ('host', '=', False)])
+        if available_slots > 0 or nb_pending == 0:
+            return
+
+        builds_to_kill = self.env['runbot.build']
+        btk = Build
```
Xavier-Do: this line looks useless

runbot/models/repo.py
```diff
@@ -550,6 +543,36 @@ def _gc_running(self, host):
...
+                newer_candidates = Build.search([
+                    ('id', '>', build.id),
+                    ('branch_id', '=', build.branch_id.id),
+                    ('build_type', '=', 'normal'),
+                    ('parent_id', '=', False),
+                    ('hidden', '=', False),
+                    ('config_id', '=', build._get_top_parent().config_id.id)
+                ])
+                if newer_candidates:
+                    btk = build if not build.parent_id else build._get_top_parent()
```
Xavier-Do: why not btk = build._get_top_parent()

runbot/models/repo.py
```diff
@@ -550,6 +543,36 @@ def _gc_running(self, host):
...
+                    btk._log('repo._update_git', 'Build automatically killed, newer build found.', level='WARNING')
+                    builds_to_kill |= btk
+        btk.write({'requested_action': 'deathrow'})
```
Xavier-Do: We can have a pending build being marked as deathrow if pending is host-defined. 
We need to skip them in this case instead.

runbot/models/repo.py
```diff
@@ -544,6 +537,40 @@ def _gc_running(self, host):
...
+                    ('config_id', '=', build._get_top_parent().config_id.id)
+                ])
+                if newer_candidates:
+                    btk = build._get_top_parent()
+                    if btk.local_state == 'pending':
+                        builds_to_skip |= btk
+                    else:
+                        btk._log('repo._update_git', 'Build automatically killed, newer build found.', level='WARNING')
+                        builds_to_kill |= btk
+        builds_to_kill.requested_action = 'deathrow'
```
Xavier-Do: Actually, writing deathrow will only kill current_build 
A solution would be to call _ask_kill but it will log stuffs. Usefull logic is mainly to manage childrens but duplicate logic is also usefull if a child of top_parent_build is a duplicate. But in this case we don't want to kill build.duplicate_id, only skip in every case. It is maybe something we don't want to do anymore in ask_kill too, we need to think about that. 

We can either:
- call _ask_kill by adapting behaviour/ logs (on both builds_to_kill and builds_to_skip, no distinction)
- copy most of the logic, maybe with something simillar to previous newer build found
- only kill builds on host without using _get_top_parent. It isn't a real problem but we may have strange behaviour: impossible to rebuild a build even if it's partially killed, red waiting build even if there are no errors. 

I think that we may want to slightly rethink _ask_kill and use that.
suggestion for _ask_kill:
-> only kill duplicate if one is a pr with pull_head_name equivalent to the other, and maybe add an option for this case
-> allow to give a kill reason that will replace default 'requested by', by a message 'Newer build found' in this case. 

Then we can try to use _ask_kill as much as possible when it makes sence.

runbot/models/build.py
```diff
@@ -902,25 +905,32 @@ def _kill(self, result=None):
...
-    def _ask_kill(self, lock=True):
+    def _is_pull_duplicate(self):
+        """ check that the build and its duplicate are related PR/branch"""
+        self.ensure_one()
+        if not self.duplicate_id:
+            return False
+        pr = self.branch_id.name.startswith('refs/pull/') and self.branch_id or self.duplicate_id.branch_id.name.startswith('refs/pull/') and self.duplicate_id.branch_id or None
+        br = self.branch_id.name.startswith('refs/heads/') and self.branch_id or self.duplicate_id.branch_id.name.startswith('refs/heads/') and self.duplicate_id.branch_id or None
+        if pr is not None and br is not None:
+            return br.branch_name == pr.pull_head_name.split(':')[-1]
```
Xavier-Do: return self.branch_name == self.duplicate_id.pull_head_name.split(':')[-1] or self.duplicate_id.branch_name == self.pull_head_name.split(':')[-1]

runbot/models/build.py
```diff
@@ -902,25 +905,32 @@ def _kill(self, result=None):
             build._github_status()
             self.invalidate_cache()
 
-    def _ask_kill(self, lock=True):
+    def _is_pull_duplicate(self):
+        """ check that the build and its duplicate are related PR/branch"""
+        self.ensure_one()
+        if not self.duplicate_id:
+            return False
+        pr = self.branch_id.name.startswith('refs/pull/') and self.branch_id or self.duplicate_id.branch_id.name.startswith('refs/pull/') and self.duplicate_id.branch_id or None
```
Xavier-Do: why not some pythonicodoo stuff

pr = next(pr for pr in (self | self.duplicate_id) if pr.pull_head_name)
br = next(br for br in (self | self.duplicate_id) if not br.pull_head_name)

runbot/models/build.py
Xavier-Do: even better:

`self.pull_branch_name == self.duplicate_id.pull_branch_name`

in all cases (in worst case will kill same pull or same branch)

PR #287, opened 2020-01-13
Master mergebot migration

xmo-odoo:
Since the runbot was migrated we might as well migrate those two as well.

### DIFF
--- a/forwardport/data/security.xml
+++ b/forwardport/data/security.xml
@@ -17,7 +17,33 @@
         <field name="perm_write">1</field>
         <field name="perm_unlink">1</field>
     </record>
+    <record id="access_forwardport_tagging_admin" model="ir.model.access">
+        <field name="name">Admin access to tagging</field>
+        <field name="model_id" ref="model_forwardport_tagging"/>
+        <field name="group_id" ref="runbot_merge.group_admin"/>
+        <field name="perm_read">1</field>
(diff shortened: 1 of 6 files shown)


xmo-odoo:
@julienlegros my local migration script is
```sh
CMD="migration/odoo-bin -c odoo/odoo.conf -d mergebot-migrated --addons-path=runbot,migration/addons,migration/odoo/addons --stop-after-init --log-handler=odoo.modules.module:CRITICAL"

dropdb --if-exists mergebot-migrated
createdb -T mergebot mergebot-migrated
# apparently mergebot's default_user is 2055 but that user was removed
psql mergebot-migrated -c "update ir_model_data set res_id=5 where module='base' and name='default_user'"

git -C migration checkout odoo/12.0
eval $CMD -ubase

git -C migration checkout odoo/13.0
eval $CMD -ubase
```
so the updates from this PR were apparently not necessary to perform the migration itself, though a -u on runbot_merge and forwardport after updating their code would be useful for the ACL and the conversion of a selection field to an integer.

xmo-odoo:
@KangOl @Xavier-Do migration script question: if I add a pre-migration script for 13.0.0.0 in runbot_merge is it going to get run during the transition from 12 to 13?

KangOl:
Yes, it will.

PR #289, opened 2020-01-14
[FIX] runbot: temporarily freeze the coverage version

d-fence:
As the Odoo requirements were recently updated, the last RUN entry in
the Docker image was rebuilt on our runbot instances.

Since that moment the coverage builds are failing with an import error.
After investigation, it appears that the latest coverage version modifies the sys.path.
A bug report was written for coverage: https://github.com/nedbat/coveragepy/issues/919

For that reason, this commit freeze the coverage version until this bug
is fixed.

Also, the os import problem in conatainer.py is fixed.

### DIFF
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -14,7 +14,7 @@ import datetime
 import io
 import json
 import logging
-from .common import os
+import os
 import shutil
 import subprocess
 import time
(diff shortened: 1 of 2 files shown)


runbot/container.py
```diff
@@ -14,7 +14,10 @@
 import io
 import json
 import logging
-from .common import os
+try:
+    from .common import os
+except ImportError:
+    import os
```
xmo-odoo: runbot.common doesn't seem to patch os, so why even try to import .common.os?

d-fence: hmmm, you are right. I think that it's an old remain of the tests refactoring.
I remove that

PR #314, opened 2020-02-10
13.0 bunch of improvements xmo

xmo-odoo:

PR #316, opened 2020-02-10
[ADD] runbot_merge: force reviewers to have a Github login

beledouxdenis:
Vincent Schippefilt (vsc) was set as reviewer
but had no Github login. How could he r+ then?

Besides, the Github provisioning would have removed
his reviewer rights automatically as his Github login
was not set / was not part of the Github logins list
of our employees on odoo.com.

### DIFF
--- a/runbot_merge/models/res_partner.py
+++ b/runbot_merge/models/res_partner.py
@@ -10,6 +10,12 @@ class Partner(models.Model):
     delegate_reviewer = fields.Many2many('runbot_merge.pull_requests')
     formatted_email = fields.Char(string="commit email", compute='_rfc5322_formatted')
 
+    _sql_constraints = [
+        ('reviewer_required_fields',
+            "CHECK(coalesce(github_login, '') != '' OR (reviewer = false AND self_reviewer = false))",
+            "A reviewer must have a Github login!"),
+    ]
+
     def _auto_init(self):
         res = super(Partner, self)._auto_init()
         tools.create_unique_index(


xmo-odoo:
Ah so I should not have added him back, probably.

Though I don't think I can merge this given I'm rewriting the review ACL in #314 and the constraint is not going to work *at all*.

beledouxdenis:
I just realized that as you just pinged me on #314. I did not know there was such a refactoring in progress. Well it should be kept in mind there should be a constraint on this. And it would be better if it was a SQL constraint in case someone adds the review rights via the SQL shell rather than the interface. This is not that unlikely as we are all a bunch of nerds who prefer the command-line rather than the gui.

xmo-odoo:
> Well it should be kept in mind there should be a constraint on this. And it would be better if it was a SQL constraint in case someone adds the review rights via the SQL shell rather than the interface. This is not that unlikely as we are all a bunch of nerds who prefer the command-line rather than the gui.

Yeah but I don't really know how to do that except a trigger and a postgres function or some such garbage.

beledouxdenis:
Woudln't something like this work?
```sql
CHECK(coalesce(github_login, '') != '' OR (SELECT bool_or(review or self_review) FROM res_partner_review WHERE partner_id = id GROUP by partner_id) = false)
```

beledouxdenis:
subqueries not allowed, shit :d

xmo-odoo:
> subqueries not allowed, shit :d

Yep, that's also where I got stuck. Apparently you have to use triggers and a custom pl function.

PR #317, opened 2020-02-10
[FIX] runbot: use a custom template to create db

d-fence:
Since cb05f2b9d8, when creating a database, the template1 is used, this
allows to customize the template and install some needed Postgress
extensions.

Unfortunately, it's also a source of build failures. For example, if the
pg_activity util is used on a runbot host, database creation may fail
with a message like this:

source database "template1" is being accessed by other users

It's because pg_activity needs a database and uses template1.

With this commit, template1 is still used by default but can be changed
with a system parameter. That way, a custom template can be created on
runbot hosts and used when creating DB in builds.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -17,6 +17,7 @@ from odoo.exceptions import UserError, ValidationError
 from odoo.http import request
 from odoo.tools import appdirs
 from collections import defaultdict
+from psycopg2 import sql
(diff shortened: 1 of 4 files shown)


Xavier-Do:
Can we bootstrap template db creation for new runbot? If not exist on worker launch, try to create db from template1. It will alwo be usefull for migration since we only need to restart workers.

PR #335, opened 2020-02-25
[FIX] runbot: displayed collapsed error

mart-e:
By default, the error was displayed with arrow pointing UP, not down

Screenshot_2020-02-25 runbot build runbot odoo com


Xavier-Do:
It broke during migration to 13.

The initial behaviour whas actually to show error message, adding 'show' class in target. What do you think?

mart-e:
Like  this?

Xavier-Do:
It was for 13.0, didn't noticed.

PR #338, opened 2020-02-26
[IMP] runbot: add coverage xml export

d-fence:
When coverage is computed, a post command is used to generate the HTML
report. In order to use the coverage result locally the HTML report is
not enough.

With this commit, an XML report is also generated. It's a single xml
file, downloadable from the build result web page.

The _post_install_command method is renamed into its plural form because
it was useless to return only one command.

### DIFF
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -380,7 +380,7 @@ class ConfigStep(models.Model):
         if extra_params:
             cmd.extend(shlex.split(extra_params))
 
-        cmd.posts.append(self._post_install_command(build, modules_to_install, py_version))  # coverage post, extra-checks, ...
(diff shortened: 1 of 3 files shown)


d-fence:
@mart-e I think this PR should solve your problem.
It will be difficult to use diff-cover during the runbot builds because it needs to be run in a git repository.
But the runbot builds use exported git archive ...

runbot/models/build_config.py
```diff
@@ -413,6 +413,10 @@ def log_end(self, build):
             kwargs['log_type'] = 'link'
         build._log('', **kwargs)
 
+        if self.coverage:
+            message = 'Coverage xml: $$fa-download'
+            build._log('end_job', message, log_type='link', path='/data/build/logs/coverage.xml')
```
Xavier-Do: Can we add a link to coverage html too, maybe using a fancy markdown log?

PR #340, opened 2020-02-28
[IMP] runbot: allow pseudo markdown in log messages

d-fence:
With this commit, a kind of markdown is allowed in log messages.
Following patterns are supported:
  **strong**
  ~~striketrough~~
  __underline__
 [link](target)
 @icon-font-awesome-class

### DIFF
--- a/runbot/models/event.py
+++ b/runbot/models/event.py
@@ -2,11 +2,12 @@
 
 import logging
 
-from odoo import models, fields, api, tools
+from ..common import pseudo_markdown
+from odoo import models, fields, tools
 
 _logger = logging.getLogger(__name__)
 
-TYPES = [(t, t.capitalize()) for t in 'client server runbot subbuild link'.split()]
+TYPES = [(t, t.capitalize()) for t in 'client server runbot subbuild link markdown'.split()]
 
 
 class runbot_event(models.Model):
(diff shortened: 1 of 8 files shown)


runbot/models/event.py
```diff
@@ -71,6 +73,43 @@ def init(self):
...
+            while p in text:
+                text = text.replace(p, b[n % 2], 1)
+                n += 1
+
+        # icons
+        re_icon = re.compile(r'@icon-(?P<icon_class>[a-zA-Z0-9-]+)')
+        res = re_icon.search(text)
+        while res:
+            icon = '<i class="fa fa-%s"></i>' % res['icon_class']
+            text = re_icon.sub(icon, text)
```
Xavier-Do: I think that something like that would be enough
```
re_icon = re.compile(r'@icon-([a-zA-Z0-9-]+)')
icon = '<i class="fa fa-\1"></i>'
ext = re_icon.sub(icon, text)
```

runbot/models/event.py
```diff
@@ -71,6 +73,43 @@ def init(self):
...
+
+        text = utils.escape(self.message)
+        patterns = {
+                '**': ('<strong>', '</strong>'),
+                '~~': ('<del>', '</del>'),  # it's not official markdown but who cares
+                '__': ('<ins>', '</ins>'),  # same here, maybe we should change the method name
+                '`': ('<code>', '</code>'),
+        }
+        n = 0
+        for p, b in patterns.items():
```
Xavier-Do:
```
re_pattern = re.compile(r'__(.+?)__')
sub = '<s>\g<1></s>'
re_pattern.sub(sub, text)

```

Xavier-Do:
The idea would also to be able to do that for child description in build view:

```
                                     <t t-if="child.description">
                                      <t t-esc="child.description" />
                                    </t>
                                    <t t-else="">
                                        with config <t t-esc="child.config_id.name"/>
                                    </t>
```

PR #342, opened 2020-03-02
[IMP] runbot: copy steps when duplicating a config

d-fence:
When a build.config is copied, the config steps are not copied.

The steps have to be explicitly ordered otherwise it leads to a
traceback when trying to copy a 'run' step which  have to be the last
one.


runbot/models/build_config.py
```diff
@@ -55,7 +55,7 @@ def unlink(self):
 
     def step_ids(self):
         self.ensure_one()
-        return [ordered_step.step_id for ordered_step in self.step_order_ids]
+        return [ordered_step.step_id for ordered_step in sorted(self.step_order_ids, key=lambda rec: rec.sequence)]
```
Xavier-Do: I may be wrong but it may be an orm bug, would be great to investigate. Except that lgtm.

PR #343, opened 2020-03-05
[FIX] runbot: fix child errors too

d-fence:
When a build error is marked as fixed, the children are left as not
fixed. If a new error arrise and is the same as one of the child, the
build error is linked to the child. The new error is the hidden in the
backend.

With this commit, when the active field is changed on the parent, the
children field is modified too.


d-fence:
pfff already fixed here 4b2a93af9e10  shame on me

PR #344, opened 2020-03-06
[IMP] runbot: bump Google Chrome version to 80

d-fence:
The fix odoo/odoo#45347 was merged two weeks ago, it's now time to
deploy on runbot.


Xavier-Do:
IHNIWIAD but LGTM

PR #345, opened 2020-03-06
[IMP] runbot: add a build stat model

d-fence:
When a build is done, various numerical informations could be extracted
from log files.  e.g.: global query count or tests query count ...

The extraction regular expression could be hard-coded in a custom step
but there is no place holder where to store the retrieved information.
In order to compare results, we need to store it.

With this commit, a new model `runbot.build.stat` is used to store
key/values pair linked to a build/config_step.  That way, extracted
values can be stored.

Also, another `runbot.build.stat.regex` is used to store regular
expressions that can be used to grep log files and extract values.

The regular expression must contain a named group like this:
`(?P<value>.+)`
The text catched by this group MUST be castable into a float.

Optionally, another named group can be used in the regular expresion
like this:
`(?P<key>.+)`
This `key` group will then be used to augment the key name in the
database.

Example:
    Consider a log line like this one:
    `odoo.addons.website_blog.tests.test_ui tested in 10.35s`

    A regular expression like this, named `test_duration`:
    `odoo.addons.(?P<key>.+) tested in (?P<value>\d+\.\d+)s`

    Should store the following key:value:
    `{
        'key': 'test_duration.website_blog.tests.test_ui',
        'value': 10.35
    }`

A `generic` boolean field is present on the build.stat.regex object,
meaning that when no regex are linked to a make_stats config step, then,
all the generic regex will be applied.

A wizard is added to help the creation the regular expressions, allowing
to test if they work against a user provided example.

A _make_stats method is added to the ConfigStep model which is called
during the _schedule of a build, just before calling the next step.
The regex search is only apllied in steps that have the `make_stats`
boolean field set to true. Also, the build branch have to be flagged
`make_stats` too or the build can have a key `make_stats` in its
config_data field.

The `make_stats` field on the branch is a compute stored field.
That way, sticky branches are automaticaly set `make_stats' true.

Finally, an SQL view is used to facilitate the stats visualisation.

### DIFF
--- a/runbot/models/branch.py
+++ b/runbot/models/branch.py
@@ -41,6 +41,8 @@ class runbot_branch(models.Model):
     branch_config_id = fields.Many2one('runbot.build.config', 'Branch Config')
     config_id = fields.Many2one('runbot.build.config', 'Run Config', compute='_compute_config_id', inverse='_inverse_config_id')
 
+    make_stats = fields.Boolean('Extract stats from logs', compute='_compute_make_stats', store=True)
+
     @api.depends('sticky', 'defined_sticky', 'target_branch_name', 'name')
     # won't be recompute if a new branch is marked as sticky or sticky is removed, but should be ok if not stored
     def _compute_closest_sticky(self):
@@ -106,6 +108,11 @@ class runbot_branch(models.Model):
     def _compute_pull_branch_name(self):
         for branch in self:
             branch.pull_branch_name = branch.pull_head_name.split(':')[-1] if branch.pull_head_name else branch.branch_name
+
+    @api.depends('sticky')
+    def _compute_make_stats(self):
+        for branch in self:
+            branch.make_stats = branch.sticky
 
     @api.depends('name')
     def _get_branch_infos(self, pull_info=None):
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -65,7 +65,7 @@ class runbot_build(models.Model):
     repo_id = fields.Many2one(related='branch_id.repo_id', readonly=True, store=True)
     name = fields.Char('Revno', required=True)
     description = fields.Char('Description', help='Informative description')
-    md_description = fields.Char(compute='_compute_md_description', String='MD Parsed Description', help='Informative description mardown parsed')
+    md_description = fields.Char(compute='_compute_md_description', String='MD Parsed Description', help='Informative description markdown parsed')
     host = fields.Char('Host')
     port = fields.Integer('Port')
     dest = fields.Char(compute='_compute_dest', type='char', string='Dest', readonly=1, store=True)
@@ -80,6 +80,7 @@ class runbot_build(models.Model):
     log_ids = fields.One2many('ir.logging', 'build_id', string='Logs')
     error_log_ids = fields.One2many('ir.logging', 'build_id', domain=[('level', 'in', ['WARNING', 'ERROR', 'CRITICAL'])], string='Error Logs')
     config_data = JsonDictField('Config Data')
+    stat_ids = fields.One2many('runbot.build.stat', 'build_id', strings='Statistics values')
 
     # state machine
 
@@ -737,6 +738,9 @@ class runbot_build(models.Model):
                 results = {'local_result': 'ko'}
 
             build_values.update(results)
+
+            # compute statistics before starting next job
+            build.active_step._make_stats(build)
 
             build.active_step.log_end(build)
 
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -102,6 +102,8 @@ class ConfigStep(models.Model):
     step_order_ids = fields.One2many('runbot.build.config.step.order', 'step_id')
     group = fields.Many2one('runbot.build.config', 'Configuration group', help="Group of config's and config steps")
     group_name = fields.Char('Group name', related='group.name')
+    make_stats = fields.Boolean('Make stats', default=False)
+    build_stat_regex_ids = fields.Many2many('runbot.build.stat.regex', string='Stats Regexes')
     # install_odoo
     create_db = fields.Boolean('Create Db', default=True, track_visibility='onchange')  # future
     custom_db_name = fields.Char('Custom Db Name', track_visibility='onchange')  # future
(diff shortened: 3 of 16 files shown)


runbot/models/build.py
```diff
@@ -729,6 +730,13 @@ def _schedule(self):
```
Xavier-Do: `if build.branch_id.sticky and build.active_step_id.make_stats and results.get('local_result') != 'ko':
`
should be an active_step logic, inside a make_stats method
`build.active_step._make_stats(build)`

and maybe we should always do the stats, no mather the result (we can always filter on build global result, not the opposite)

runbot/models/build_config.py
```diff
@@ -565,6 +566,23 @@ def _make_tests_results(self, build):
             build_values['local_result'] = build._get_worst_result([build.local_result, local_result])
         return build_values
 
+    def _make_stats(self, build):
+        build._log('make_stats', 'Getting stats from log file')
+        if (build.branch_id.sticky or build.config_data.get('make_stats')) and self.make_stats:
```
Xavier-Do: Thinking about it, should we have the opportunity to make stat on a branch? should we have default config_data on branch? or a make_state on branch without using config_data in a hardcoded step? (no answers, only question good enough to test the feature)

runbot/models/build_stat_regex.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @api.model
+    def _find_in_file(self, file_path):
+        """ Search file for all regex and returns a list of (key,value) """
+        if not os.path.exists(file_path):
+            return []
+        key_values = []
+        with open(file_path, "r") as log_file:
+            for line in log_file.readlines():
+                for build_stat in self.search([]):
```
Xavier-Do: :@ :$

runbot/models/build_config.py
```diff
@@ -565,6 +566,23 @@ def _make_tests_results(self, build):
...
+                message = '**An error occured while computing statistics of %s:**\n`%s`' % (build.job, str(e).replace('\\n', '\n').replace("\\'", "'"))
+                build._log('make_stats', message, level='INFO', log_type='markdown')
+
+    def _grep_statistics(self, build):
+        """ extract statistics from  step log file with build stat"""
+        log_path = build._path('logs', '%s.txt' % self.name)
+        if not os.path.exists(log_path):
+            return
+        key_values = self.env['runbot.build.stat.regex']._find_in_file(log_path)
+        self.env['runbot.build.stat']._write_key_values(build, self, key_values)
```
Xavier-Do: either rename _grep_statistics, or move _write_key_values in make stats.

I think that in the future we may add a many2many between steps and stats.regex. So a good apy would be 
```
self.env['runbot.build.stat.regex'].search([])._find_in_file(log_path)
```

hidding statistic

runbot/models/build_stat_regex.py
d-fence: :scream: Ah, you find it, it was a trap

runbot/models/build_stat.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+    config_step_name = fields.Char(String="Config Step name", readonly=True)
+    build_id = fields.Many2one("runbot.build", string="Build", readonly=True)
+    build_name = fields.Char(String="Build name", readonly=True)
+    branch_id = fields.Many2one("runbot.branch", string="Branch", readonly=True)
+    branch_name = fields.Char(string="Branch name", readonly=True)
+    branch_sticky = fields.Boolean(string="Sticky", readonly=True)
+    repo_id = fields.Many2one("runbot.repo", string="Repo", readonly=True)
+    repo_name = fields.Char(string="Repo name", readonly=True)
+
```
Xavier-Do: just some suggestion that could be interresting:
host -> aggregate on host to check if it is slower
build_config_id -> coverage/flamegraph/other comparaison.
build_parent_path

runbot/models/build_config.py
```diff
@@ -565,6 +567,19 @@ def _make_tests_results(self, build):
...
         return build_values
 
+    def _make_stats(self, build):
+        build._log('make_stats', 'Getting stats from log file')
+        log_path = build._path('logs', '%s.txt' % self.name)
+        if not os.path.exists(log_path):
+            return
+        if (build.branch_id.make_stats or build.branch_id.sticky or build.config_data.get('make_stats')) and self.make_stats:
+            try:
+                key_values = self.build_stat_regex_ids._find_in_file(log_path)
```
Xavier-Do: suggestion:
```
regex_ids = self.build_stat_regex_ids
if not regex_ids:
    regex_ids = regex_ids.search(['!', ('name', '=like', '_%')])
key_values = regex_ids._find_in_file(log_path)

```

Xavier-Do: maybe cleaner: add an active field, and add a context={'active_test': False} context on build_stat_regex_ids field

runbot/models/branch.py
```diff
@@ -41,6 +41,8 @@ class runbot_branch(models.Model):
     branch_config_id = fields.Many2one('runbot.build.config', 'Branch Config')
     config_id = fields.Many2one('runbot.build.config', 'Run Config', compute='_compute_config_id', inverse='_inverse_config_id')
 
+    make_stats = fields.Boolean('Extract stats from logs', default=False)
```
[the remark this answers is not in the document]
d-fence: > It can be done latter if you don"t want to change it now.

No, I'll do it

runbot/models/branch.py
```diff
@@ -41,6 +41,8 @@ class runbot_branch(models.Model):
     branch_config_id = fields.Many2one('runbot.build.config', 'Branch Config')
     config_id = fields.Many2one('runbot.build.config', 'Run Config', compute='_compute_config_id', inverse='_inverse_config_id')
 
+    make_stats = fields.Boolean('Extract stats from logs', compute='_compute_make_stats', store=True, default=True)
```
Xavier-Do: default is nof usefull with compute, and anyway should be default to False

runbot/models/build_config.py
```diff
@@ -102,6 +102,8 @@ class ConfigStep(models.Model):
     step_order_ids = fields.One2many('runbot.build.config.step.order', 'step_id')
     group = fields.Many2one('runbot.build.config', 'Configuration group', help="Group of config's and config steps")
     group_name = fields.Char('Group name', related='group.name')
+    make_stats = fields.Boolean('Make stats', default=False)
+    build_stat_regex_ids = fields.Many2many('runbot.build.stat.regex', string='Stats Regexes', context={'active_test': False})
```
Xavier-Do: active_test context  is not usefull anymore

d-fence: good catch

PR #346, opened 2020-03-09
[FIX] runbot_merge: fix forward-port Odoo wiki link

d-fence:


xmo-odoo:
@d-fence did someone un-rename this page?

d-fence:
I duno. It was reported that the link was broken.

xmo-odoo:
Looks like it was @mart-e

mart-e:
Yes because @rim-odoo broke links by renaming the page before, just restored the good old way

rim-odoo:
sorry guys ><

xmo-odoo:
@mart-e yeah but we'd fixed everything to match so you just broke it a second time.

PR #353, opened 2020-03-30
[FIX] runbot: use proxy-mode for running instances

d-fence:
When deployed behind nginx reverse proxy and https, the Odoo proxy-mode
must be enabled.


julienlegros:
:+1:

PR #356, opened 2020-04-06
[FIX] runbot: send status only once.

Xavier-Do:
Runbot can send status multiple time for the same hash:
- if transaction fails in scheduler and is retried
- if multiple subbuild are failing

Leading to multiple issues:
- when github receive more than one failure status, mergebot will
be notified multiple times and send multiple mail (for forward ports mainly)
- github will answer `422 Unprocessable Entity for url...` after
1000 status.

This fix proposes to limit number of status:
- By avoiding to send status for orphan build (parent status will never change)
- By storing last send status to avoid to notify multiple time
- By sending status post commit to avoid to contact  github in case of failure.
This will also slightly reduce transaction time by removing an http request.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -12,7 +12,7 @@ from ..common import dt2time, fqdn, now, grep, local_pgadmin_cursor, s2human, Co
 from ..container import docker_build, docker_stop, docker_state, Command
 from ..fields import JsonDictField
 from odoo.addons.runbot.models.repo import RunbotException
-from odoo import models, fields, api
+from odoo import models, fields, api, registry
 from odoo.exceptions import UserError, ValidationError
 from odoo.http import request
 from odoo.tools import appdirs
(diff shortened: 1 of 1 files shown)


runbot/models/build.py
```diff
@@ -1040,19 +1041,40 @@ def _github_status_notify_all(self, status):
...
-                except Exception:
-                    self._log('_github_status_notify_all', 'Status notification failed for "%s" in repo "%s"' % (self.name, repo.name))
+            if status != self.last_github_status:
+                repos_ids = {b.repo_id.id for b in self.search([('name', '=', self.name)])}
+                build_name = self.name
+                user_id = self.env.user.id
+                _dbname = self.env.cr.dbname
+                _context = self.env.context
+                build_id = self.id
+                self.last_github_status = status
```
d-fence, line 1051: status is a dict containing various information like 'target_url' and 'description'.
We should only compare the important keys: 'state' and 'context'

runbot/models/build.py
```diff
@@ -1040,19 +1041,41 @@ def _github_status_notify_all(self, status):
...
         self.ensure_one()
         if self.config_id.update_github_state:
-            repos = {b.repo_id for b in self.search([('name', '=', self.name)])}
-            for repo in repos:
-                _logger.debug("github updating %s status %s to %s in repo %s", status['context'], self.name, status['state'], repo.name)
-                try:
-                    repo._github('/repos/:owner/:repo/statuses/%s' % self.name, status, ignore_errors=True)
-                except Exception:
-                    self._log('_github_status_notify_all', 'Status notification failed for "%s" in repo "%s"' % (self.name, repo.name))
+            state = status.get('state')
```
d-fence, line 1044: if `legal/cla` is sent first with state `success`, it means that `ci/runbot/` with same sate will not be sent, right ?

Xavier-Do: check moved to _github_status

PR #359, opened 2020-04-08
[IMP] runbot: add unaccent option in config file

d-fence:
In 5bba39e15394d unaccent was enabled in container tests, it's now time
to enable it in runbot code.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -1034,6 +1034,9 @@ class runbot_build(models.Model):
                 os.mkdir(datadir)
             command.add_config_tuple("data_dir", '/data/build/datadir')
 
+        if grep(config_path, "--unaccent"):
+            command.add_config_tuple('unaccent', 'True')
+
         return command
 
     def _github_status_notify_all(self, status):


d-fence:
@xmo-odoo @Xavier-Do 
Do not yet merge.
Not confident about this one.

runbot/models/build.py
```diff
@@ -1034,6 +1034,9 @@ def _cmd(self, python_params=None, py_version=None, local_only=True, sub_command
                 os.mkdir(datadir)
             command.add_config_tuple("data_dir", '/data/build/datadir')
 
+        if grep(config_path, "--unaccent"):
+            command.add_config_tuple('unaccent', 'True')
```
d-fence, line 1038: I see two potential problems here:
1. We [create db localy](https://github.com/odoo/runbot/blob/49c1d960fa36b543395a9106aac9e0c77e63be30/runbot/models/build_config.py#L355) before launching Odoo and I think that the unaccent option is used at db creation in Odoo so this maybe useless
2. When testing locally, if the unaccent extension is not available, Odoo logs an Error and this may be a problem for other runbot users

Xavier-Do: For the second point we don't really car.

But I think that the first plan for unaccent was to use another template. Couldn't we juste add the extension to runbot_template? Another solution is to let odoo create the database.

d-fence: If the only purpose of '--unaccent' is to install the 'unaccent' extension, the present PR is probably useless as it's already installed in template1.

Xavier-Do: I think it is, or at least it was.

PR #370, opened 2020-05-16
[FIX] runbot: handle git archive failure

d-fence:
When a failure occurs during the export of the commit, if the failure is
at the git command level the error message is lost in the system logs.

At the user level, the only feedback is that no server was found in the
sources, which is a bit obscure.

With this commit, the return code of the git process is verified and an
Exception is raised in case of failure.


Xavier-Do:
This was actually manually added in new runbot

PR #379, opened 2020-06-18
[FIX] runbot: sanitize build dest also

d-fence:
In a previous commit 9684c2d, the docker name was sanitized to prevent
crash when the docker name contains unallowed characters.

Unfortunately, some characters are able to create problems elsewhere.
For example, when a database is downloaded, if a `#` is present in the
branch name, the zip filename will contains this character too, leading
to 404 when trying to download the zip.

To avoid further problems, this fix use the function from the previous
commit to sanitize the build dest, as many other names are computed from
the dest.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -9,7 +9,7 @@ import subprocess
 import time
 import datetime
 from ..common import dt2time, fqdn, now, grep, local_pgadmin_cursor, s2human, Commit, dest_reg, os, list_local_dbs, pseudo_markdown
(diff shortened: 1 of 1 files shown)


Xavier-Do:
Shouldn't we remove the docker sanitize ton only cleanup dest?

d-fence:
> Shouldn't we remove the docker sanitize ton only cleanup dest?

I think that the Docker name sanitization  should stay, and as you stated that this fix is temporary I made this small fix.

d-fence:
not needed anymore

PR #383, opened 2020-07-08
[FIX] runbot: mis-migrated test

xmo-odoo:
73f720a55c8d2c3b3f867e293ff5cb6fb93281a7 refactored the runbot tests, and amongst other things created a single patch point for the "mock root" as a testcase attribute.

One of the tests was missed during that refactoring, likely because it's skipped by default.


d-fence:
Nice catch, thanks

xmo-odoo:
> Nice catch, thanks

I'm completely innocent, pylint caught it while I was looking at something completely unrelated.

PR #384, opened 2020-07-14
Make statuses more flexible

xmo-odoo:
* split statuses into their own object which can be filtered on a per-branch basis
* further allow statuses to be filtered on being required for PRs or stagings (or both, or none though that's not very useful)
* add support for overriding statuses


xmo-odoo:
@Xavier-Do I guess I'll update the mergebot next monday?

PR #393, opened 2020-09-01
[FIX] runbot: set default result before github

obayit:
Set the default 'ok' to local_result before calling github.
This will prevent making all builds fail when testing isn't enabled.

We only use runbot to make sure there is no "hard" errors, also testing is a resource intensive task.
Currently all our builds are marked as failed in github. This did fix it for me.


Xavier-Do:
Hello. We actually did the exact same fix on our side so I guess I can close this.

PR #395, opened 2020-09-10
[IMP] runbot: add bundle_ids to ErrorLog SQL view

d-fence:


runbot/models/event.py
```diff
@@ -125,11 +125,59 @@ def action_goto_build(self):
...
+                SELECT * FROM runbot_batch_slot as slot
+                JOIN
+                  runbot_batch batch ON batch.id = slot.batch_id
+                JOIN
+                  runbot_bundle bundle ON bundle.id = batch.bundle_id
+                WHERE
+                  bundle.sticky = %s AND slot.build_id = build.id);
+            """, (value,))
+            build_ids = [t[0] for t in self.env.cr.fetchall()]
+            return [('top_parent_id', 'in', build_ids)]
```
Xavier-Do, line 174: This doesn't scale very well.

Xavier-Do (commented):
I need to do some more testing bug the 7 days filter solution will only work because there is not to much result, so the search_count is not triggered (if the search returns less result than limit, the search_count is not triggered)

The limit is actually only used for pagination.
So, if there is only one page, the search_count will be skipped, and if there is more than one page we actually don't care so much if the result is not right.

I propose:

 ```
def search_count(self, args):
    return 4242
```
From my experiments performances look ok, we just don't have the perfect result as count but who cares for this model.

d-fence:
> I need to do some more testing bug the 7 days filter solution will only work because there is not to much result, so the search_count is not triggered (if the search returns less result than limit, the search_count is not triggered)
> 
> The limit is actually only used for pagination.
> So, if there is only one page, the search_count will be skipped, and if there is more than one page we actually don't care so much if the result is not right.
> 
> I propose:
> 
> ```python
> def search_count(self, args):
>    return 4242
> ```
> 
> From my experiments performances look ok, we just don't have the perfect result as count but who cares for this model.

I updated as proposed. Quickly tested locally and works as expected.

PR #400, opened 2020-09-14
[IMP] runbot: don't reserve host when there is only one

obayit:
When using only a single host, if it is reserved, then builds will not start.

### DIFF
--- a/runbot/models/host.py
+++ b/runbot/models/host.py
@@ -107,5 +107,5 @@ class Host(models.Model):
         self.ensure_one()
         nb_hosts = self.env['runbot.host'].search_count([])
         nb_reserved = self.env['runbot.host'].search_count([('assigned_only', '=', True)])
-        if nb_reserved < (nb_hosts / 2):
+        if nb_reserved < (nb_hosts / 2) and nb_hosts > 1:
             self.assigned_only = True


Xavier-Do:
I have to admit that out /2 condition is mainly based on runbot.odoo.com use cases. The main reason is that if a git repo is corrupted (with 30 host running we have at least one per month) we don't want this worker to fail all builds: export fails leading to quick red build. A host with a corrupted git is a red build factory.  A host should only be reserved after a git error on a testing build. Is it a problem in this case to reserve the host? All your build may fail in this case. Can you describe how you got this problem?

Except that I have no problem with your suggested changes.

obayit:
"Can you describe how you got this problem?"
I have runbot installed in only one host, when a git error occures (then the host gets reserved), the one and only host never gets assigned any new build.
Thus I have to fix or disable the failing repo, then go to the `runbot.host` and change the `assigned_only` to false. Finally I created a custom module and overriden the `disable` method to never set `assigned_only`, since I know that I don't have multiple hosts.
The suggested change is for when there is only one host, with multiple repositories to test. If one repository failed, the host should _ideally_ be assigend builds from other repositories.

Xavier-Do:
I was sure I answered but I may have forgot to submit. 

> I have runbot installed in only one host, when a git error occures (then the host gets reserved), the one and only host never gets assigned any new build.

It is intended as broken test says ;) 

> Thus I have to fix or disable the failing repo, then go to the `runbot.host` and change the `assigned_only` to false. Finally I created a custom module and overriden the `disable` method to never set `assigned_only`, since I know that I don't have multiple hosts.

Your custom module may be a solution but the **initial problem is the failing repo. This is what should be fixed**. Build will be red if the repo cannot be fetched. Since you have a local fix I don't consider this as a priority, we can still talk about that.

> The suggested change is for when there is only one host, with multiple repositories to test. 

I actually think that the **nb_reserved < (nb_hosts / 2)** is already to much. We never had a problem with to many host being reserved. This was a security if a repo was invalid, to avoid to block all other builds.

> If one repository failed, the host should _ideally_ be assigned builds from other repositories.

This is a good idea but way to complex to implement for the small benefit.

Maybe we can fond the reason why it is a problem for you. For us, we need this check in case a repo is temporary in failure (IO/network error on one host). If it fails for you it is more likely a configuration problem I guess? Maybe the problem is just that it is not obvious enough that the host was reserved? Maybe we could add a warning on fronted when to many host are reserved? Maybe we could check if the repo is valid before saving it? My question "Can you describe how you got this problem" was more "Why did you get a git error"

obayit:
> Why did you get a git error

I tried to not answer that question, but it appears there is no way around it :)

I got the error for multiple reasons:
  - I added a repo that the runbot user which executes `git` doesn't have permissions on
  - I added a repo  in a `gitlab` server that is not stable (the server fails after every restart).

In all cases runbot stops working. And I had to unreserve the host manually evertime.

> If it fails for you it is more likely a configuration problem I guess?

Yes it is a configuration problem, but I figured it may encouter many others.

> Maybe the problem is just that it is not obvious enough that the host was reserved?

No, the problem is that there is only one host, that means it should never be reserved. And just get assigned builds even if it means they will inevitably fail.

> Maybe we could check if the repo is valid before saving it?

Good idea, but what if it got invalid after it was working fine? myabe it could be disabled, and then send a notification after.

> Maybe we could add a warning on fronted when to many host are reserved?

Another good idea.

Xavier-Do:
I understand the problem in your use case, but it doesn't apply to everyone and most of the time the auto reserve behavior is interesting to keep even with one host. Since you have an external module fixing this issue, I guess it works and it is enough for you? If someone else have the same problem I'll do something about that. (maybe an icp).

obayit:
Totally agree with you.
I thought this would apply to everyone else, but apparently it is not!

~~What is an icp? if it is a configuration option, I think it will just add to the complexity :(

PR #403, opened 2020-09-15
[IMP] runbot: add pdf417 library

d-fence:
The pdf417 python library is needed for the l10n_cl_edi. As this module
is not mandatory, the library was not added in the requirements.txt
file.

In order to test the feature on the runbot, this commit adds the library
in the Docker container.

Related work:
    odoo/odoo#54995
    odoo/enterprise#12080
    odoo/enterprise@8290bfaf


Xavier-Do (approved):
Ok I guess :D

d-fence:
@Xavier-Do I'm just waiting the confirmation that it works with the 0.8.0 version before merging and deploying

PR #411, opened 2020-10-10
[IMP] runbot: add extra parameters and environment variables to build config step

davejames:
Currently there is no way that I can see to configure extra parameters or environment variables to run_odoo build config steps. It is possible on install steps, but not on run steps. This PR adds that capability.

This is useful in cases where a module needs to be run as a server wide module. For example [OCA/queue](https://github.com/OCA/queue/tree/13.0/queue_job).

### DIFF
--- a/runbot/views/config_views.xml
+++ b/runbot/views/config_views.xml
@@ -72,9 +72,9 @@
                         <field name="test_tags"/>
                         <field name="enable_auto_tags"/>
                         <field name="sub_command"/>
-                        <field name="extra_params" groups="base.group_no_one"/>
                     </group>
-                    <group string="Test settings" attrs="{'invisible': [('job_type', 'not in', ('python', 'install_odoo', 'test_upgrade'))]}">
+                    <group string="Extra Parameters" attrs="{'invisible': [('job_type', 'not in', ('python', 'install_odoo', 'test_upgrade', 'run_odoo'))]}">
+                        <field name="extra_params"/>
                         <field name="additionnal_env"/>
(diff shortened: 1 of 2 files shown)


runbot/views/config_views.xml
```diff
@@ -62,6 +62,13 @@
                         <field name="running_job"/>
                         <field name="ignore_triggered_result"/>
                     </group>
+                    <group string="Runtime Parameters"
+                           attrs="{'invisible': [('job_type', '!=', ('run_odoo'))]}"
```
Xavier-Do, line 66: To clean it a little, you can remove the `groups="base.group_no_one"`. We tried to hide some  magic fields but it was a bad idea.

I also think we can avoid to duplicate extra_params field and KISS by merging this with the second badly copy/pasted _Test setting_ group and adding  `run_odoo` to `('python', 'install_odoo', 'test_upgrade')` as well as moving extra_params to this section. It makes sence to add extra_params support for test_upgrade steps too if we need it. You can rename this section in "Extra Parameters" (or something else if you have a better idea)

Xavier-Do (changes requested):
Small remarks but except that LGTM. FYI we are planning to create a 14.0 branch "soon".

davejames:
@Xavier-Do can you please let me know what error is being reported by runbot? This is a private build so I don't have access, but I can't see anything that might have caused the build to fail

Xavier-Do:
It looks like I broke 13.0 Friday, you can rebase it is now fixed  ;)

PR #412, opened 2020-10-11
[IMP] runbot: add a model to handle multiple Dockerfile

d-fence:
As of now, runbot uses a single Dockerfile maintained in runbot code.
Despite the advantage of having a single file to maintain, it limits the testing possibilities.

For example, we would like to test Odoo versions against different Ubuntu/Debian versions.
Also, maintaining the Dockerfile in the source code enforce the maintainers to update runbot instances each time the Dockerfile is updated. e.g.: testing different chrome versions.

This PR is an attempt to use a Dockerfile model composed of steps (the same way as config steps), so a Dockerfile can be built from different step blocks.

Odoo versions can be linked to a different Dockerfile and a bundle can override  the version Dockerfile.


runbot/container.py
```diff
@@ -270,6 +268,23 @@ def _docker_ps():
     return output.strip().split('\n')
 
 
+def docker_list_images():
```
Xavier-Do, line 271: This is not used I think

runbot/data/dockerfile_data.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <record model="runbot.dockerfile" id="runbot.docker_default">
+        <field name="name">Docker Default</field>
+        <field name="template_id" ref="runbot.docker_base"/>
+        <field name="to_build">True</field>
+        <field name="description">Default Dockerfile with Python3.
+Can be used with 12 to 14.
```
Xavier-Do, line 8: "from" instead of "with" maybe. Anyway, I wonder if it is a good idea to add version number in the code, this will add the need to maintain this line when releasing a new version even if the docker_default doesn"t change.

runbot/models/host.py
```diff
@@ -71,10 +71,23 @@ def _bootstrap(self):
...
-        docker_build(log_path, static_path)
+        self.clear_caches()  # needed to ensure that content is updated on all hosts
+        for dockerfile in self.env['runbot.dockerfile'].search([('to_build', '=', True)]):
+            _logger.info('Building %s, %s', dockerfile.name, hash(dockerfile.dockerfile))
+            docker_build_path = os.path.join(static_path, 'docker', dockerfile.image_tag)
+            os.makedirs(docker_build_path, exist_ok=True)
+            with open(os.path.join(docker_build_path, 'Dockerfile'), 'w') as Dockerfile:
+                Dockerfile.write(dockerfile.dockerfile)
+            build_process = docker_build(docker_build_path, dockerfile.image_tag)
+            if build_process.returncode !=0:
```
Xavier-Do, line 85: Lint

Xavier-Do (commented):
Small commantes and LGTM

PR #413, opened 2020-10-12
[FIX] runbot: serve badges for non-default projects

davejames:
presently, regardless of the trigger or repo passed in the URL, the project is hardcoded and the returned badge will always correspond to a bundle in the main project.

This PR determines the project from either the trigger or the repo passed in the URL, which results in the correct badge being returned.


Xavier-Do:
Thanks for the fix, good job!

PR #424, opened 2020-11-13
[IMP] runbot: allow 'main' as alternative to 'master' branch name

gurneyalex:
github is proposing "main" as a replacement to "master" for the default branch name. This PR allows runbot to consider both as equivalent.

### DIFF
--- a/runbot/data/runbot_data.xml
+++ b/runbot/data/runbot_data.xml
@@ -29,7 +29,7 @@
 
     <record model="ir.config_parameter" id="runbot.runbot_is_base_regex">
         <field name="key">runbot.runbot_is_base_regex</field>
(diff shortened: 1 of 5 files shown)


Xavier-Do:
This is a bit dangerous for us because is someone push a "main" branch this will be considered as an equivalent to master witch is not the case for us. More than that you're not managing all case such as batch prepare fallback on master for commit filling. 

I think that an acceptable change would be to change 'master' occurences by an optionnal 'master_name' icp, so that we can opt-in to use "main" instead of "master" if needed. Managing both could lead to stange behaviours. runbot_is_base_regex can be edited the same way if needed.

gurneyalex:
My main concern is that new repostories don't get a master branch on github, so we may run into issues there. Maybe it would be simpler if Odoo followed this huge move and renamed master to main on its repository.

PR #425, opened 2020-11-13
[IMP] runbot: improve coverage

Xavier-Do:
- execute coverage results in case of failure
- fix coverage links
- use coverage directly


d-fence:
LGTM.
Can be merged if you think it's ready.

PR #446, opened 2020-12-30
[FIX] crash when github returns an HTTP Error

gurneyalex:
Github returns a 403 when a service has reached the max api limit -> handle this error,
and possibly other HTTP errors to avoid the exception reaching the cron


Xavier-Do:
Hello.

This is intended: We always assume that we have pr info in most computes about pr. If not, things could go wrong. 
When configured with a valid token, github should always answer exept when some service are down. In this case the transaction is rollbacked and runbot will try again untill success.

If it is a problem, you can still disable pr fetching on a remote and only branches will be used. Ci should appear on the pr anyway since it is commit based.

PR #447, opened 2021-01-01
[FIX] runbot: bug in duplicate module detection

gurneyalex:
The test in the original code will never fire because the value searched for is not in the keys of the dictionary, but in one of the lists which are in the values. Work around this by maintaining a reverse dictionary module name -> commit and use this for the test.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -765,18 +765,23 @@ class BuildResult(models.Model):
         return exports
 
     def _get_available_modules(self):
+        all_modules = dict()
         available_modules = defaultdict(list)
         # repo_modules = []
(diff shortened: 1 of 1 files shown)


runbot/models/build.py
```diff
@@ -765,18 +765,23 @@ def _checkout(self):
         return exports
 
     def _get_available_modules(self):
+        all_modules = dict{}
```
Xavier-Do, line 768: Invalid syntax ;)

Except that, LGTM

Xavier-Do (changes requested):

runbot/models/build.py
gurneyalex: gah....

PR #423, opened 2020-11-13
[FIX] runbot bundle missing base_id

gurneyalex:
When trying to set up my runbot instance, I ran into an issue where the base_id of a bundle
was sometimes empty. I'm not sure exactly what causes this, possibly it happens when
bootstrapping a runbot from scratch, or there is an issue with the triggers of the computed
field. The net result was that my PR branches had no base_id in their bundles and would not build.

This PR fixes the issue by explicitly calling the _compute_base_id method when the bundle has no version
or no base id.

PR #448, opened 2021-01-03
[FIX] runbot.bundle compute base_id

gurneyalex:
the matching done on the bundle name was fragile:

* PR comming from clones have a prefix before the branch name
* Some PR can use a slightly different nameing convention (e.g. forget
the ".0" in the version number, use a different separator than '-'
* we should try really hard to find a good version, and using the target
branch name as a fallback sounds like a good idea

### DIFF
--- a/runbot/models/bundle.py
+++ b/runbot/models/bundle.py
@@ -1,6 +1,7 @@
 import time
 import logging
 import datetime
+import re
 import subprocess
 
 from collections import defaultdict
@@ -93,6 +94,13 @@ class Bundle(models.Model):
                 continue
             project_id = bundle.project_id.id
             master_base = False
+            bundle_name = bundle.name
+            if ':' in bundle_name:
+                bundle_name = bundle_name.split(':', 1)[1]
+            if re.match(r'^\d\d[-_]', bundle_name):
+                bundle_name = bundle_name[:2] + '.0-' + bundle_name[4:]
+            elif re.match(r'^\d\d\.\d_', bundle_name):
+                bundle_name = bundle_name[:4] + '-' + bundle_name[6:]
             for bid, bname in self._get_base_ids(project_id):
                 if bundle.name.startswith('%s-' % bname):
                     bundle.base_id = self.browse(bid)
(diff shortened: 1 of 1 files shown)


gurneyalex:
This may be the cause for the issue I try to tackle in #423

gurneyalex:
unfortunately I cannot check the CI failure (403 - Unauthorized)

Xavier-Do:
Sorry for the long response time, I'll try to follow this one. 

Here is the error message

2021-01-03 07:44:47,585 32 INFO 5592420-13-0-runbot_coverage odoo.modules.loading: loading runbot/data/runbot_data.xml 
2021-01-03 07:44:47,626 32 WARNING 5592420-13-0-runbot_coverage odoo.addons.runbot.models.bundle: Could not guess the branch_id of bundle Dummy. Checking the target branch 

I didn't really took the time to understand the cause of the problem yet at first glance this fix looks a little too complex:

I may be missing something but it looks like you are not using `bundle_name`?

> PR comming from clones have a prefix before the branch name

This should be managed correctly by runbot, they are considered external pr and the target_branch_name may be used by compute_bundle_id to define the base when creating a new bundle for this pr.

> Some PR can use a slightly different nameing convention (e.g. forget  the ".0" in the version number, use a different separator than '-'

Runbot asks for strict convention when naming branches in guidelines. Runbot is based on this guideline and prefers to fail that trying to guess. This is a good way to enforce this convention.Anyway, if no base is found master will be used by default and can be changed manually on a bundle by setting "defined_base_id" if needed to avoid to create a new branch.

> we should try really hard to find a good version, and using the target branch name as a fallback sounds like a good idea

Yes it is, but most of the time when the target repo of the pr is known by runbot, the bundle should already exists since the branch is created before the pr. If not, the base_id is set manually by _compute_bundle_id (defined_base_id) since it isn't the default branch name case. This part is quite dirty and could be improved but we need to think about all cases (creation order, retarget, manual choice, ...)

Xavier-Do:
Making some cleanup in the old forgotten pull request but this won't be merged:

This is not compatible with existing and future changes, we want to enforce a precise branch naming. External branches works fine with pr (using target).

Usually if something does not go well to define the base in my experience, it is because the base does not exist in the repo. Other issue where fixed in the past, like a very old commit on a base branch, no master branch at all, ....

PR #455, opened 2021-02-08
[IMP] runbot: add an option to exclude paths from coverage report

davejames:
As a developer of modules on top of Odoo, I would like to exclude Odoo's files from the coverage report as I am more interested in the coverage of my own modules. By having Odoo's files listed alongside my own, the total coverage score is diluted.

This commit adds a field to the build config step to enter a comma separated list of paths to omit from coverage.

I'm not sure if it will be of any utility to Odoo SA, but will be useful for all the module developers out there and does not get in the way if not used.

### DIFF
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -127,6 +127,7 @@ class ConfigStep(models.Model):
     db_name = fields.Char('Db Name', compute='_compute_db_name', inverse='_inverse_db_name', tracking=True)
     cpu_limit = fields.Integer('Cpu limit', default=3600, tracking=True)
     coverage = fields.Boolean('Coverage', default=False, tracking=True)
+    paths_to_omit = fields.Char('Paths to omit from coverage', tracking=True)
     flamegraph = fields.Boolean('Allow Flamegraph', default=False, tracking=True)
     test_enable = fields.Boolean('Test enable', default=True, tracking=True)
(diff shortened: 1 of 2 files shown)


runbot/models/build_config.py
```diff
@@ -830,8 +831,11 @@ def _post_install_commands(self, build, modules_to_install, py_version=None):
     def _perfs_data_path(self, ext='log'):
         return '/data/build/logs/flame_%s.%s' % (self.name, ext)
 
-    def _coverage_params(self, build, modules_to_install):
-        pattern_to_omit = set()
+    def _coverage_params(self, build, modules_to_install, paths_to_omit=None):
+        if paths_to_omit:
+            pattern_to_omit = set(paths_to_omit.split(','))
+        else:
+            pattern_to_omit = set()
```
Xavier-Do, line 838: Hello. 

Why not using self.path_to_omit here? 

```suggestion
    def _coverage_params(self, build, modules_to_install, paths_to_omit=None):
        pattern_to_omit = set()
        if self.paths_to_omit:
            pattern_to_omit = set(self.paths_to_omit.split(','))
```

Xavier-Do (changes requested):

runbot/models/build_config.py
davejames: good idea, I have revised as you suggested

Xavier-Do:
@d-fence this will need a update on next deploy.

PR #491, opened 2021-06-10
[IMP] runbot: add a pr_state field on bundle

d-fence:
In order to facilitate bundle filtering, a new computed field is added
that gives a state for all the PR in a bundle. This state will give an
information about the `task` state.

If all the PR in a bundle are closed, the pr_state is considered done,
on the other hand, if at least one PR is still open, the whole bundle is
considered open from the `task`point of view. A third state (`nopr`) is
used for bundles that does not have any PR.


d-fence:
@Xavier-Do split the pr_state feature from #490 
I applied the recommendations from you review

d-fence:
Change to a simple stored boolean field has_open_pr

PR #496, opened 2021-07-02
[IMP] use random order of module installation

yelizariev:
Some errors could be caught only on installing modules in specific order. It's
better to see them randomly than to don't see the errors at all


Xavier-Do:
Runbot is trying to have a determinist behaviour.  If we want to do that, it should be done in a python config_step, and by fixing the seed in order to be able to reproduce the error if something breaks.

Anyway odoo is sorting the module from the -i so it shouldn't change anything in term of execution, except that the used command line will be harder to read ;)

PR #502, opened 2021-07-09
[IMP] runbot: display dates with timezone

d-fence:


runbot/templates/commit.xml
```diff
@@ -70,7 +70,7 @@
             <table class="table table-sm table-borderless">
               <tr t-foreach='last_status_by_context' t-as='context'>
                 <t t-set="status" t-value="last_status_by_context[context]"/>
-                <td t-esc="status.sent_date and status.sent_date.strftime('%Y-%m-%d %H:%M:%S') or '—'"/>
+                <td><span t-if="status.sent_date" t-field="status.sent_date" t-options="{'format': 'yyyy-MM-dd HH:mm:ssZ'}"/><t t-else="">&amp;mdash;</t></td>
```
Xavier-Do, line 73: why not adding %z in strftime ?

d-fence: datetime are naïve on python side

d-fence:
to redo in V14

PR #512, opened 2021-07-29
[FIX] runbot: correctly get number type value from icp

kos94ok-3D:
If set `containers_memory_limit = 2.0`
In row:

> 
> containers_memory_limit = self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_containers_memory', 0)
>  kwargs['memory'] = containers_memory_limit * 1024 ** 3

I get `"2.02.02.02.02.02.02.02.02.02.02.02.02.02.0...."`

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -736,7 +736,7 @@ class BuildResult(models.Model):
             self._log('Preparing', 'Using Dockerfile Tag %s' % kwargs['image_tag'])
         containers_memory_limit = self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_containers_memory', 0)
         if containers_memory_limit and 'memory' not in kwargs:
-            kwargs['memory'] = containers_memory_limit * 1024 ** 3
+            kwargs['memory'] = float(containers_memory_limit) * 1024 ** 3
         docker_run(**kwargs)
 
     def _path(self, *l, **kw):


Xavier-Do:
Hello. 
Thanks for your contribution
Can you please adapt your commit message to avoid a line break?
Also you don't need to specify [13.0] in the commit message ;)

[FIX] runbot: correctly get number type value from icp

runbot/models/build_config.py
```diff
@@ -242,7 +242,7 @@ def _run(self, build):
         self._run_step(build, log_path)
 
     def _run_step(self, build, log_path, **kwargs):
-        build.log_counter = self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_maxlogs', 100)
+        build.log_counter = int(self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_maxlogs', 100))
```
Xavier-Do, line 245: The Integer fields will implicitelly cast to int so this is not needed ;) Please remove this diff
https://github.com/odoo/odoo/blob/13.0/odoo/fields.py#L1201

runbot/models/host.py
```diff
@@ -20,7 +20,7 @@ class Host(models.Model):
     assigned_only = fields.Boolean('Only accept assigned build', default=False, tracking=True)
     nb_worker = fields.Integer(
         'Number of max paralel build',
-        default=lambda self: self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_workers', default=2),
+        default=lambda self: int(self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_workers', default=2)),
```
Xavier-Do, line 23: Same as log_counter, this is not needed.

runbot/models/build.py
```diff
@@ -734,7 +734,7 @@ def _docker_run(self, **kwargs):
             kwargs.update({'image_tag': self.params_id.dockerfile_id.image_tag})
         if kwargs['image_tag'] != 'odoo:DockerDefault':
             self._log('Preparing', 'Using Dockerfile Tag %s' % kwargs['image_tag'])
-        containers_memory_limit = self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_containers_memory', 0)
+        containers_memory_limit = float(self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_containers_memory', 0))
```
Xavier-Do, line 737: Good catch! But can you add a cast to int on the `kwargs['memory'] = containers_memory_limit * 1024 ** 3`  line?

Xavier-Do (changes requested):

runbot/models/build.py
```diff
@@ -736,7 +736,7 @@ def _docker_run(self, **kwargs):
             self._log('Preparing', 'Using Dockerfile Tag %s' % kwargs['image_tag'])
         containers_memory_limit = self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_containers_memory', 0)
         if containers_memory_limit and 'memory' not in kwargs:
-            kwargs['memory'] = containers_memory_limit * 1024 ** 3
+            kwargs['memory'] = float(containers_memory_limit) * 1024 ** 3
```
Xavier-Do, line 739: My last suggestion was mainly to add a cast to int afterward, I add no problem with the cast to float on the previous lines ;)
```suggestion
            kwargs['memory'] = int(float(containers_memory_limit) * 1024 ** 3)
            
```
It's working now but this would be safer in case docker makes change on the command line. Except that LGTM

Xavier-Do:
Closed at 5b9c20877e86e7e0f554a65ff95a672a6e330f1e

PR #532, opened 2021-09-10
[FIX] runbot: allow a runbot admin to create runbot batch log

d-fence:
When a runbot admin creates a cron for e.g. nightly builds, the cron may
fails if a runbot.batch.log is written.

With this commit, a runbot admin can create such a log.


Williambraecky (approved):
Rebased for 18.0

PR #558, opened 2021-11-25
[IMP] forwardport: add missing space

mart-e:
Instead of "[FW]Create README.md", the forwardport PR is now named "[FW] Create README.md"


xmo-odoo:
A very common case is that the source PR is formatted like a commit though, in that case this change will result in e.g. 

    [FW] [FIX] point_of_sale : Positive tax for credit note

rather than the current

    [FW][FIX] point_of_sale : Positive tax for credit note

which I think is detrimental

PR #566, opened 2021-12-28
[FIX] runbot: set version_id for new base bundles

### LINKED ISSUE #565: [BUG] `runbot.branch` creates a base bundle without version
In this [line](https://github.com/odoo/runbot/blob/d8b96db1b709bd3e3b11718b5bb39f763d922429/runbot/models/branch.py#L159) when a base bundle is created it doesn't set the version, which causes an error later on when creating a new build (`psycopg2.errors.NotNullViolation: null value in column "version_id" violates not-null constraint...`).

obayit:
In new installations of runbot this is necessary, otherwise the user will have to edit the base bundles manually to set the `version_id`.
I am not sure if there will be any unwanted side effects from this, e.g: creating versions for unnecessarily.

Closes #565

### DIFF
--- a/runbot/models/branch.py
+++ b/runbot/models/branch.py
@@ -157,6 +157,7 @@ class Branch(models.Model):
                 }
                 if need_new_base:
                     values['is_base'] = True
+                    values['version_id'] = self.env['runbot.version']._get(name)
 
                 if branch.is_pr and branch.target_branch_name:  # most likely external_pr, use target as version
                     base = self.env['runbot.bundle'].search([


obayit:
```
    @api.depends('is_base', 'base_id.version_id')
    def _compute_version_id(self):
        for bundle in self.sorted(key='is_base', reverse=True):
            if not bundle.is_base:
                bundle.version_id = bundle.base_id.version_id
                continue
            bundle.version_id = self.env['runbot.version']._get(bundle.name)
```
The version should be created by this compute function.

PR #575, opened 2022-01-25
first fix runbot module: docker build now works

yustas147:
without this fix default docker image could not be built


Xavier-Do:
This is fixed in ec2f12ef5f28099a6ee1084b43a21a0508bd996c

PR #583, opened 2022-02-07
[IMP] runbot: fully garbage collect old builds

d-fence:
When a build is garbage collected and the local folders are cleaned, the
log files are kept to allow build errors investigations.

With this commit, a full local cleanup will be done 365 days after the
garbage collect.  This means that the build dir will be completely
removed.  The default number of days can be tweaked with the
`runbot.full_gc_days` ir.config_parameter.

Also, the _local_cleanup method can be called with a `full` boolean
parameter to force a full cleanup. e.g.: when called in a config step.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -393,7 +393,6 @@ class BuildResult(models.Model):
     @api.depends_context('batch')
     def _compute_build_url(self):
         batch = self.env.context.get('batch')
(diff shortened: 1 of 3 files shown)


Xavier-Do (commented):
I would just add the icp to res.config.settings and maybe gc the *run.txt and *wake_up.txt in all cases

PR #596, opened 2022-03-15
[IMP] runbot: commit exports 

Xavier-Do:
Source cleanup will check in multiple place for potentially used
commits. This maked the cleanup logic complex, plus limit the usable
commits in python steps. The current use case is to export the mergebase
commits.

The poposed solution is to save the exported commit and clean this list
when the build is done. All commit in this list cannot be cleaned.

### DIFF
--- a/runbot/models/commit.py
+++ b/runbot/models/commit.py
@@ -52,11 +52,11 @@ class Commit(models.Model):
                     module = os.path.basename(os.path.dirname(manifest_path))
                     yield (addons_path, module, manifest_file_name)
 
-    def export(self):
+    def export(self, build):
         """Export a git repo into a sources"""
         #  TODO add automated tests
         self.ensure_one()
-
+        self.env['runbot.commit.export'].create({'commit_id': self.id, 'build_id': build.id})
         export_path = self._source_path()
 
         if os.path.isdir(export_path):
@@ -228,3 +228,13 @@ class CommitStatus(models.Model):
                 self._cr.after('commit', send_github_status_async)
             else:
                 send_github_status(self.env)
+
+
+class CommitExport(models.Model):
+    _name = 'runbot.commit.export'
+    _description = 'Commit export'
+
+    build_id = fields.Many2one('runbot.build', index=True)
+    commit_id = fields.Many2one('runbot.commit')
+
+    host = fields.Char(related='build_id.host', store=True)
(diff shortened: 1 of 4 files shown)


runbot/models/runbot.py
```diff
@@ -315,10 +315,15 @@ def _source_cleanup(self):
             if self.pool._init:
                 return
             _logger.info('Source cleaning')
-            # we can remove a source only if no build are using them as name or rependency_ids aka as commit
+
+            cannot_be_deleted_path = set()
+            for commit in self.env['runbot.commit.export'].search([('host', '=', fqdn())]).mapped('commit_id'):
```
d-fence, line 320:
```suggestion
            for commit in self.env['runbot.commit.export'].search([('build_id.host', '=', fqdn())]).mapped('commit_id'):
```

runbot/models/commit.py
```diff
@@ -228,3 +228,13 @@ def send_github_status_async():
...
             else:
                 send_github_status(self.env)
+
+
+class CommitExport(models.Model):
+    _name = 'runbot.commit.export'
+    _description = 'Commit export'
+
+    build_id = fields.Many2one('runbot.build', index=True)
+    commit = fields.Many2one('runbot.commit')
```
d-fence, line 238:
```suggestion
    commit_id = fields.Many2one('runbot.commit')
```

d-fence (changes requested):

runbot/models/runbot.py
Xavier-Do: `self.env['runbot.commit.export'].search([('host', '=', '')])`                                                                                                               

> SELECT "runbot_commit_export".id FROM "runbot_commit_export" WHERE ("runbot_commit_export"."host" = '') ORDER BY "runbot_commit_export"."id" 

vs

`self.env['runbot.commit.export'].search([('build_id.host', '=', 'runbot31.odoo.com')])`
                                   
>SELECT "runbot_build".id FROM "runbot_build" WHERE ("runbot_build"."host" = 'runbot31.odoo.com') ORDER BY "runbot_build"."id"   
> SELECT "runbot_commit_export".id FROM "runbot_commit_export" WHERE id in **{insert 365891 ids here }** ORDER BY "runbot_commit_export"."id"

PR #597, opened 2022-03-18
[IMP] runbot: use docker python lib

d-fence:
While using the docker cli as a subprocess was KISS and convenient,
the python Docker SDK is mature, easy to use, available as a Debian
package and much more powerful.

It will permit to monitor the containers memory consumption and will
help to spot memory leaks.

### DIFF
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -191,7 +191,7 @@ def _docker_run(cmd=False, log_path=False, build_dir=False, container_name=False
         for dp, hp in enumerate(exposed_ports, start=8069):
             ports[f'{dp}/tcp'] = ('127.0.0.1', hp)
 
-    ulimits = []
+    ulimits = [docker.types.Ulimit(name='core', soft=0, hard=0)]  # avoid core dump in containers
     if cpu_limit:
         ulimits.append(docker.types.Ulimit(name='cpu', soft=cpu_limit, hard=cpu_limit))
 
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -10,14 +10,24 @@ When testing this file:
 """
 import configparser
 import io
-import json
 import logging
 import os
 import re
 import subprocess
+import warnings
+
+# unsolved issue https://github.com/docker/docker-py/issues/2928
+with warnings.catch_warnings():
+    warnings.filterwarnings(
+        "ignore",
+        message="The distutils package is deprecated.*",
+        category=DeprecationWarning
+    )
+    import docker
 
 
 _logger = logging.getLogger(__name__)
+
 DOCKERUSER = """
 RUN groupadd -g %(group_id)s odoo \\
 && useradd -u %(user_id)s -g odoo -G audio,video odoo \\
@@ -103,14 +113,23 @@ def _docker_build(build_dir, image_tag):
     """Build the docker image
     :param build_dir: the build directory that contains Dockerfile.
     :param image_tag: name used to tag the resulting docker image
+    :return: tuple(success, msg) where success is a boolean and msg is the error message or None
     """
     # synchronise the current user with the odoo user inside the Dockerfile
     with open(os.path.join(build_dir, 'Dockerfile'), 'a') as df:
         df.write(DOCKERUSER)
-    log_path = os.path.join(build_dir, 'docker_build.txt')
-    logs = open(log_path, 'w')
-    dbuild = subprocess.Popen(['docker', 'build', '--tag', image_tag, '.'], stdout=logs, stderr=logs, cwd=build_dir)
-    return dbuild.wait()
+    docker_client = docker.from_env()
+    try:
+        docker_client.images.build(path=build_dir, tag=image_tag, rm=True)
+    except docker.errors.APIError as e:
+        _logger.error('Build of image %s failed with this API error:', image_tag)
+        return (False, e.explanation)
+    except docker.errors.BuildError as e:
+        _logger.error('Build of image %s failed with this BUILD error:', image_tag)
+        msg = f"{e.msg}\n{''.join(l.get('stream') or '' for l in e.build_log)}"
+        return (False, msg)
+    _logger.info('Dockerfile %s finished build', image_tag)
+    return (True, None)
 
 
 def docker_run(*args, **kwargs):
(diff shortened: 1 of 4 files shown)


runbot/container.py
```diff
@@ -195,18 +211,24 @@ def _docker_stop(container_name, build_dir):
...
     _logger.info('Stopping container %s', container_name)
+    docker_client = docker.from_env()
+    container = docker_client.containers.get(container_name)
     if build_dir:
         end_file = os.path.join(build_dir, 'end-%s' % container_name)
         subprocess.run(['touch', end_file])
     else:
         _logger.info('Stopping docker without defined build_dir')
-    subprocess.run(['docker', 'stop', container_name])
+    container.stop()
```
Xavier-Do, line 221: We may want to set a timeout to 1
We may want to catch APIError (maybe if the docker isn't running anymore)

runbot/container.py
```diff
@@ -195,18 +211,24 @@ def _docker_stop(container_name, build_dir):
...
 def docker_is_running(container_name):
     container_name = sanitize_container_name(container_name)
-    dinspect = subprocess.run(['docker', 'container', 'inspect', container_name], stderr=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
-    return True if dinspect.returncode == 0 else False
+    docker_client = docker.from_env()
+    try:
+        container = docker_client.containers.get(container_name)
+    except docker.errors.NotFound:
+        return False
+    return True if container.status == 'running' else False
```
Xavier-Do, line 231:
```suggestion
    return container.status == 'running'
```

runbot/container.py
```diff
@@ -103,14 +105,27 @@ def _docker_build(build_dir, image_tag):
...
     # synchronise the current user with the odoo user inside the Dockerfile
     with open(os.path.join(build_dir, 'Dockerfile'), 'a') as df:
         df.write(DOCKERUSER)
-    log_path = os.path.join(build_dir, 'docker_build.txt')
-    logs = open(log_path, 'w')
-    dbuild = subprocess.Popen(['docker', 'build', '--tag', image_tag, '.'], stdout=logs, stderr=logs, cwd=build_dir)
-    return dbuild.wait()
+    docker_client = docker.from_env()
+    try:
+        docker_client.images.build(path=build_dir, tag=image_tag)
```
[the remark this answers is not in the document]
d-fence: After some rumination, I think that I prefer to keep the plain file.
My actual use case is that if it fails, it's far easier to repeat the docker build manually on the runbot where it failed.
You just change to the docker build dir and repeat the `docker build .` command.
With the file obj, you will have dump the dockerfile first, save it somewhere, apply the patch ...

runbot/container.py
```diff
@@ -103,14 +113,27 @@ def _docker_build(build_dir, image_tag):
...
+        return (False, e.explanation)
+    except docker.errors.BuildError as e:
+        _logger.error('Build of image %s failed with this BUILD error:', image_tag)
+        _logger.error('   "%s"', e.msg)
+        with tempfile.NamedTemporaryFile(mode='w',prefix='docker_build_log', delete=False, dir=build_dir) as logfile:
+            for log_line in e.build_log:
+                logfile.write(log_line.get('stream') or '')
+        return (False, e.msg)
+    _logger.info('Dockerfile %s finished build', image_tag)
+    return (True, None)
```
Xavier-Do, line 136:
```suggestion
    except docker.errors.APIError as e:
        _logger.error('Build of image %s failed with this API error:', image_tag)
        return (False, e.explanation)
    except docker.errors.BuildError as e:
        _logger.error('Build of image %s failed with this BUILD error:', image_tag)
        message = e.msg + ''.join(log_line.get('stream') or '' for log_line in e.build_log)
        return (False, message)
    _logger.info('Dockerfile %s finished build', image_tag)
    return (True, None)
```

runbot/container.py
```diff
@@ -146,26 +169,16 @@ def _docker_run(cmd=False, log_path=False, build_dir=False, container_name=False
...
 
     if ro_volumes:
         for dest, source in ro_volumes.items():
             logs.write("Adding readonly volume '%s' pointing to %s \n" % (dest, source))
-            docker_command.append('--volume=%s:/data/build/%s:ro' % (source, dest))
-
-    if env_variables:
-        for var in env_variables:
-            docker_command.append('-e=%s' % var)
+            volumes.update({source: {'bind': f'/data/build/{dest}', 'mode': 'ro'}})
```
Xavier-Do, line 181:
```suggestion
            volumes[source] = {'bind': f'/data/build/{dest}', 'mode': 'ro'}
```

runbot/container.py
```diff
@@ -174,15 +187,34 @@ def _docker_run(cmd=False, log_path=False, build_dir=False, container_name=False
...
+        mem_limit=memory,
+        ports=ports,
+        ulimits=ulimits,
+        environment=env_variables,
+        init=True,
+        command=['/bin/bash', '-c',
+                 f'exec &>> /data/buildlogs.txt ;{run_cmd}'],
+        auto_remove=True,
+        detach=True
+    )
```
Xavier-Do, line 217: One part of the  "file based docker state" is that in some cases, even if the container was started, the returned status was that it wasn't (asking to fast after the docker run)

Hi would propose to test that with the docker api to check if it may be possible to simplify that.


```suggestion
    container = docker_client.containers.run(
        image_tag,
        name=container_name,
        volumes=volumes,
        shm_size='128m',
        mem_limit=memory,
        ports=ports,
        ulimits=ulimits,
        environment=env_variables,
        init=True,
        command=['/bin/bash', '-c',
                 f'exec &>> /data/buildlogs.txt ;{run_cmd}'],
        auto_remove=True,
        detach=True
    )
    if container.status != 'running':
            _logger.error('Container started but status is not running:  %s' % container.status)
```

Xavier-Do (changes requested):

runbot/container.py
```diff
@@ -195,18 +227,28 @@ def _docker_stop(container_name, build_dir):
...
 def docker_is_running(container_name):
     container_name = sanitize_container_name(container_name)
-    dinspect = subprocess.run(['docker', 'container', 'inspect', container_name], stderr=subprocess.DEVNULL, stdout=subprocess.DEVNULL)
-    return True if dinspect.returncode == 0 else False
+    docker_client = docker.from_env()
+    try:
+        container = docker_client.containers.get(container_name)
+    except docker.errors.NotFound:
+        return False
+    return container.status == 'running'
```
Xavier-Do, line 251: Me may consider it running if it is in the 'paused' state ( and maybe created or restarting)

PR #599, opened 2022-03-23
[IMP] runbot: adapt for odoo 15.0

Xavier-Do:


Xavier-Do:
Not fully tested yet but should already be a good start

PR #626, opened 2022-07-05
[FIX] runbot: avoid sudo when using pip

d-fence:
In bd4cf76b7 a new argument `--progress-bar off` was added to the pip
command line. Unfortunately, it appears that this argument is not
available in some pip versions, typically the one from Ubuntu Bionic
(9.0.1) [0]. The argument appeared in version 10.0.0 [1].

Also, in the DockerFile, we allow the user to use `sudo` for
`/usr/bin/pip3` and the main reason for that was to avoid having `pip`
to re-install the package that are pre-installed by the distribution
package manager.

That said, it appears that since pip 10.0.0 when installing packages
locally for the user, pip now properly detects distribution installed
packages.

So, as the solution to fix the progress bar issue is to upgrade pip to a
newer version, we can take advantage of it to get rid of the `sudo`.

Besides the fact that `--user` is the default on Debian based
distributions [2] , we enforce it in case another distribution is used
in the Dockerfile.

[0](https://packages.ubuntu.com/bionic-updates/python3-pip)
[1](https://pip.pypa.io/en/stable/news/#id744)
[2](https://wiki.debian.org/Python#Deviations_from_upstream)

### DIFF
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -32,9 +32,7 @@ DOCKERUSER = """
 RUN groupadd -g %(group_id)s odoo \\
 && useradd -u %(user_id)s -g odoo -G audio,video odoo \\
 && mkdir /home/odoo \\
-&& chown -R odoo:odoo /home/odoo \\
-&& echo "odoo ALL= NOPASSWD: /usr/bin/pip" > /etc/sudoers.d/pip \\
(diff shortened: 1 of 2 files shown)


d-fence:
Looks good to me but it implies to update old Dockerfiles on runbot in order to upgrade pip
Dockerfile <= Bionic

PR #638, opened 2022-07-29
[IMP] runbot: let the build insert logs into local database

d-fence:
When the number of runbot builder is growing, the concurency for inserting logs increase drastically.
This PR is an attempt to mitigate this issue by allowing runbot builds to insert logs into a local database instead of the one from the runbot leader.
After each loop turn, it's now the builder that will insert the logs directly into the leader database.

### DIFF
--- a/runbot/models/host.py
+++ b/runbot/models/host.py
@@ -1,12 +1,17 @@
 import logging
 import getpass
+
+from collections import defaultdict
+
 from odoo import models, fields, api
 from odoo.tools import config
-from ..common import fqdn, local_pgadmin_cursor, os
+from ..common import fqdn, local_pgadmin_cursor, os, list_local_dbs, local_pg_cursor
 from ..container import docker_build
+
 _logger = logging.getLogger(__name__)
 
 forced_host_name = None
+
 
 class Host(models.Model):
     _name = 'runbot.host'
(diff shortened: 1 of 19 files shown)


runbot/models/host.py
```diff
@@ -123,3 +169,56 @@ def disable(self):
...
+                    ir_log['level'] = 'SEPARATOR'
+                    ir_log['func'] = ''
+                    ir_log['type'] = 'runbot'
+                    ir_log['message'] = 'Log limit reached (full logs are still available in the log file)'
+                elif build.log_counter < 0:
+                    continue
+                if ir_log['level'].upper() == 'WARNING':
+                    build.triggered_result = 'warn'
+                elif ir_log['level'].upper() == 'ERROR':
+                    build.triggered_result = 'ko'
```
Xavier-Do, line 215: We should override the `ir.logging.create` to add this logic as said earlier. 

the code build._log('Some message', level="error") should automatically pass the build in error.

This code should be create multi, and batch by builds. 

It is a little sad to reproduce the same logic as here to group log by build but this may be necessary except if we keep a simple logic in build._log but this would be a duplicate anyway.

We should play with  _get_worst_result and only try to update the result if the result changed to avoid useless concurrent updates on global_result.

runbot/models/host.py
```diff
@@ -123,3 +169,56 @@ def disable(self):
...
+                ir_log['build_id'] = build.id
+                logs_to_send.append({k:ir_log[k] for k in ir_log if k != 'id'})
+
+        if logs_to_send:
+            self.env['ir.logging'].create(logs_to_send)
+
+        if local_log_ids:
+            with local_logs_cursor() as local_cr:
+                local_cr.execute("DELETE FROM ir_logging WHERE build_id in %s", local_log_ids)
```
Xavier-Do, line 224: This is unlikely to fail and is done in another transaction.
The insert is done in the current transaction but is not committed. It could fail while trying a global_result. We may need to commit before dropping the existing ir_logging. I just don't see where process_logs is called right now to see if it would be problematic.

runbot/models/host.py
```diff
@@ -1,12 +1,33 @@
...
+def local_logs_cursor():
+    cnx = None
+    try:
+        cnx = psycopg2.connect(f"dbname={LOGSDB}")
+        yield cnx.cursor()
+    finally:
+        if cnx:
+            cnx.close()
+
+
```
Xavier-Do, line 30: I wonder if we may prefer to keep an open cnx on this table.

Xavier-Do (changes requested):
I don't see any call to process_logs

I think we should have a commit inside before deleting the existing logs.
This means that we may want to call that quite low, in the _scheduler

I would put that just before `_get_builds_to_schedule` so that we have much chances that the list is empty when processing a build.

We could also add a last check inside the def `_schedule(` to be sure that there is no log at all 
before ending the build. Just before 

```
build_values = {
                'job_end': now(),
                'docker_start': False,
            }
```
If there is any remaining log for this build, just return so that we manage the end at the next iteration after the process_log call. We shouldn't call process_log there to avoid a commit at this level. We shouldn't have any log remaining there 99% of the time but lets be safe.

runbot/models/host.py
Xavier-Do: Also, not sure if we should set triggered_result or local_result.

Triggered_result won't trigger the global result recompute..
but triggered result was mainly usefull for the trigger to avoid concurrent update there, and this is still a write on the build anyway.

Xavier-Do:
Batching build is good for fetch but actually bad for concurrent updates. I wonder

PR #641, opened 2022-08-09
[IMP] runbot: add sort and filter capability on the build error page

rdeodoo:
Before this commit, the build errors page was neither sortable neither
filterable.

This commit adds a way to filter by:
- all
- unassigned
- seen more than once

It also allows to sort by:
- last seen
- nb seen

Typical need is to sort by nb seen and filter out the only seen once to
be able to figure which one is supposed to be checked in priority.

See result: https://watch.screencastify.com/v/L3x0qnuSgZOMQUjYDKk3

### DIFF
--- a/runbot/__manifest__.py
+++ b/runbot/__manifest__.py
@@ -2,7 +2,7 @@
 {
     'name': "runbot",
     'summary': "Runbot",
-    'description': "Runbot for Odoo 13.0",
+    'description': "Runbot for Odoo 15.0",
     'author': "Odoo SA",
     'website': "http://runbot.odoo.com",
     'category': 'Website',
--- a/runbot/controllers/frontend.py
+++ b/runbot/controllers/frontend.py
@@ -7,7 +7,7 @@ import functools
 import werkzeug.utils
 import werkzeug.urls
 
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
 from werkzeug.exceptions import NotFound, Forbidden
 
 from odoo.addons.http_routing.models.ir_http import slug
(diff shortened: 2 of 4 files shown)


rdeodoo:
@odoo/rd-runbot Hello :wave: 
That's the need we have I told you about: being able to sort and filter out some errors.
Thought I would give you a hand and suggest a change.

runbot/__manifest__.py
```diff
@@ -2,7 +2,7 @@
 {
     'name': "runbot",
     'summary': "Runbot",
-    'description': "Runbot for Odoo 13.0",
+    'description': "Runbot for Odoo 15.0",
```
rdeodoo, line 5: Or I am missing something? Tried to make it run on 13.0 but..

runbot/templates/build_error.xml
```diff
@@ -113,9 +113,15 @@
                   <t t-set="dashboard" t-value="team.dashboard_id"/>
                 </t>
               </div>
-              <h3 t-if="team.build_error_ids">Team assigned Errors</h3>
+              <div class="d-flex">
```
rdeodoo, line 116: This `<div>` is obviously not stable friendly in regular repo, is the same policy applied in this repo?
I could find another way to make the searchbar looks decent without modifying existing DOM if needed I guess.

d-fence (changes requested):
Thanks for this nice improvement. A small fix and it will be LGTM

runbot/controllers/frontend.py
```diff
@@ -431,10 +432,33 @@ def build_errors(self, sort=None, page=1, limit=20, **kwargs):
     @route(['/runbot/teams', '/runbot/teams/<model("runbot.team"):team>',], type='http', auth='user', website=True, sitemap=False)
     def team_dashboards(self, team=None, hide_empty=False, **kwargs):
         teams = request.env['runbot.team'].search([]) if not team else None
+        domain = [('id', 'in', team.build_error_ids.ids)]
```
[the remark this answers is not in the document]
rdeodoo: Changed :+1: (maybe we could have split the controllers / views for those 2 cases (teams vs team) but I let you decide as I am not very familiar with what can be done in stable in runbot anyway)

Xavier-Do:
Great work thanks!

PR #642, opened 2022-08-11
[FIX] runbot: mitigate docker status race condition

d-fence:
In some rare cases, a docker container has a status of exited, removed
or in removal and the `end-`file has been written right after the code
checked the existence of the file.

To mitigate the issue, this commit checks the `end-` file existence
after the container status have been checked. That way, if the file
exists we can be pretty confident that the build really ended.


runbot/container.py
```diff
@@ -259,6 +258,8 @@ def docker_state(container_name, build_dir):
             state = 'RUNNING' if container.status in ('created', 'running', 'paused') else 'GHOST'
         except docker.errors.NotFound:
             state = 'GHOST'
+        if state == 'GHOST' and os.path.exists(os.path.join(build_dir, f'end-{container_name}')):
```
Xavier-Do, line 261: we may want to add a comment

PR #657, opened 2022-08-30
[IMP] runbot: multi-versions branch support

xavieralt:
Some repositories host branches with addons that should be tested
on multiples versions, so to ease those use cases, this commit:

* Adds a "Multi-Versions Branch" flag on runbot.project, to indicate
  that we should not rely on the branch name to determine the version
  of the target builds.

  When set, we also hide version on related bundles/batches as only
  version/commit on the build are relevant.

* Adds a "Target Version" on runbot.trigger and runbot.bundle.trigger.custom,
  that (if set) overwrite the base bundle used to determine the missing
  commits of a specific build

* Adds a "Dockerfile" on runbot.trigger, allowing forcing specific
  image to use for a trigger.

### DIFF
--- a/runbot/models/batch.py
+++ b/runbot/models/batch.py
@@ -305,18 +305,58 @@ class Batch(models.Model):
             config = trigger_custom.config_id if trigger_custom else trigger.config_id
             if not config:
                 continue
+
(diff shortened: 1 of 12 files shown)


Xavier-Do:
I don't think this is useful anymore, feel free to reopen if I missed something :)

PR #659, opened 2022-08-31
[IMP] runbot: display assigned fixer on build page

d-fence:


d-fence:
Already done here b606a2199bf

PR #695, opened 2022-12-08
[IMP] runbot: add cpus parameter

d-fence:
When a build eats all the CPU's resources, it may interfere with other builds and cause collateral damages.

With this commit, a new settings parameter `Containers CPUs` is added in order to limit the usage of available CPU's on runbot instances.

If left to 0, no limts are applied.

Otherwise, the cpu_quota docker parameter is computed as Containers CPU's * (logical cpu's count / nb parallel builds) * cpu period which defaults to 100000.

e.g.:
- on a host with 16 logical CPU's
- with 8 parallel builds allowed
- with Containers CPUs set to 1.5
- with the default cpu_period cpu_quota will be:
    (16/8) * 1.5 * 100000 = 300000

This system parameter can be overridden by the `container_cpus` field on
steps.

references: 
  *  https://docs.docker.com/config/containers/resource_constraints/#cpu
  * https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.run

### DIFF
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -269,7 +269,6 @@ class ConfigStep(models.Model):
                     raise UserError('Invalid extra_params on config step')
 
     def _run(self, build):
-        log_path = build._path('logs', '%s.txt' % self.name)
         build.write({'job_start': now(), 'job_end': False})  # state, ...
         log_link = ''
         if self._has_log():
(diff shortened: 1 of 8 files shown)


d-fence:
Note that with the removal of the `log_path` parameter, some custom python step will have to be updated.
e.g.:  `test_migration_pre`

Xavier-Do (changes requested):

runbot/models/build_config.py
```diff
@@ -267,24 +269,34 @@ def _check(self, values):
...
-    def _run_step(self, build, log_path, **kwargs):
+    def _run_step(self, build, **kwargs):
         build.log_counter = self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_maxlogs', 100)
         run_method = getattr(self, '_run_%s' % self.job_type)
-        docker_params = run_method(build, log_path, **kwargs)
+        docker_params = run_method(build, **kwargs)
         if docker_params:
+            if 'log_path' not in docker_params:
+                docker_params['log_path'] = build._path('logs', f'{self.name}.txt')
+            if 'cpu_limit' not in docker_params:
```
Xavier-Do, line 288: todo add cpu_limit=None for _run_run_odoo

d-fence:
Done for 17.0 in #864

PR #696, opened 2022-12-08
[IMP] runbot: cross project trigger

Xavier-Do:
Allow to create a trigger without project, that execute for every project. In this case the repos are all the repo of the project that where updated.

This is mainly for psxrunbot.

Also allows to create a trigger that cross category

### DIFF
--- a/runbot/models/batch.py
+++ b/runbot/models/batch.py
@@ -162,8 +162,9 @@ class Batch(models.Model):
             _logger.error('No dockerfile found !')
 
         triggers = self.env['runbot.trigger'].search([  # could be optimised for multiple batches. Ormcached method?
-            ('project_id', '=', project.id),
-            ('category_id', '=', self.category_id.id)
+            ('category_id', 'in', [False, self.category_id.id]),
+            ('project_id', 'in', [project.id, False] if project.enable_base_triggers else [project.id]),
+            ('project_id', '=', False),
         ]).filtered(
             lambda t: not t.version_domain or \
             self.bundle_id.version_id.filtered_domain(t.get_version_domain())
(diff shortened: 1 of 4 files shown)


runbot/models/batch.py
```diff
@@ -162,8 +162,11 @@ def _prepare(self, auto_rebase=False):
             _logger.error('No dockerfile found !')
 
         triggers = self.env['runbot.trigger'].search([  # could be optimised for multiple batches. Ormcached method?
-            ('project_id', '=', project.id),
-            ('category_id', '=', self.category_id.id)
+            '|',
+            ('category_id', '=', self.category_id.id),
+            ('category_id', '=', False),
```
MiquelRForgeFlow, line 167:
```suggestion
            ('category_id', 'in', [False, self.category_id.id]),
```

:)

Xavier-Do: Thanks for the suggestion

 FYI this was just an idea we didn't tested yet because we are not sure we want this feature :) Did you test it?

MiquelRForgeFlow: No, I was just seeing the code.

runbot/models/batch.py
```diff
@@ -162,8 +162,9 @@ def _prepare(self, auto_rebase=False):
             _logger.error('No dockerfile found !')
 
         triggers = self.env['runbot.trigger'].search([  # could be optimised for multiple batches. Ormcached method?
-            ('project_id', '=', project.id),
-            ('category_id', '=', self.category_id.id)
+            ('category_id', 'in', [False, self.category_id.id]),
+            ('project_id', 'in', [project.id, False] if project.enable_base_triggers else [project.id]),
+            ('project_id', '=', False),
```
MiquelRForgeFlow, line 167:
```suggestion
            ('project_id', 'in', [project.id, False] if project.enable_base_triggers else [project.id]),
```

PR #725, opened 2023-01-17
[FIX] runbot: get_module remove prefix

Xavier-Do:
Since removeprefix was not available in ubuntu 20.04, a easier alternative using rebase was used.

The initial assumption was that the prefix would look like `odoo/addons/` and won't be in the filename.

When the repo is enterprise, the prefix is `enterprise/` meaning that module name ending with `enterprise` will be truncated

`repo._get_module('enterprise/mail_enterprise/static/src/widgets/form_renderer/form_renderer.js')`

will output

`mail_static`


Xavier-Do:
@seb-odoo FYI

We won't deploy that right now but this will be fixed sson ;)

PR #731, opened 2023-02-02
[IMP] runbot_builder: support addons_path parameter

davejames:
Currently, the addons_path is intuited from the location of the script, with the assumption that the only custom addons path will be runbot itself.

This commit introduces an addons_path parameter to support custom deployments. If the parameter is not set, the existing behaviour is unchanged.

### DIFF
--- a/runbot_builder/tools.py
+++ b/runbot_builder/tools.py
@@ -113,6 +113,7 @@ def run(client_class):
     parser.add_argument('--db_user')
     parser.add_argument('--db_password')
     parser.add_argument('-d', '--database', default='runbot', help='name of runbot db')
+    parser.add_argument('--addons-path', type=str, dest="addons_path")
     parser.add_argument('--logfile', default=False)
     parser.add_argument('--forced-host-name', default=False)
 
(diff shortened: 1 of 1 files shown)


runbot_builder/tools.py
```diff
@@ -113,6 +113,7 @@ def run(client_class):
     parser.add_argument('--db_user')
     parser.add_argument('--db_password')
     parser.add_argument('-d', '--database', default='runbot', help='name of runbot db')
+    parser.add_argument('--addons-path', type=str, nargs=1, dest="paths")
```
Xavier-Do, line 116: Not sure about this one but in any case can you rename path to addons_path ?

runbot_builder/tools.py
```diff
@@ -137,7 +138,8 @@ def run(client_class):
     odoo.tools.config['db_password'] = args.db_password
     addon_path = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), '..'))
     config_addons_path = odoo.tools.config['addons_path']
-    odoo.tools.config['addons_path'] = ','.join([config_addons_path, addon_path])
+    runbot_addons_paths = ','.join(args.paths) if args.paths else addon_path
+    odoo.tools.config['addons_path'] = ','.join([config_addons_path, runbot_addons_paths])
```
Xavier-Do, line 142: Maybe we could maybe just do something like 
`config_addons_path = args.addons_paths or odoo.tools.config['addons_path']`

runbot_builder/tools.py
davejames: yes I agree, done

runbot_builder/tools.py
davejames: @Xavier-Do Your way is cleaner, but it's easy to end up with the runbot path twice in the addons_path as the user will most likely specify the runbot path in their configuration. So I added a check to make sure that it only ends up in the addons_path once

PR #733, opened 2023-02-09
[FIX] runbot: warn when a local pg backend cannot be terminated

d-fence:
For whatever reason, it happens that we open a testing database in a psql console as another user. In that case, the runbot is unable to terminate the backend and goes in an error loop.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -19,6 +19,7 @@ from odoo.tools.safe_eval import safe_eval
 from collections import defaultdict
 from pathlib import Path
 from psycopg2 import sql
+from psycopg2.errors import InsufficientPrivilege
 import getpass
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


runbot/models/build.py
```diff
@@ -876,7 +877,11 @@ def filter_patterns(patterns, default, all):
     def _local_pg_dropdb(self, dbname):
         with local_pgadmin_cursor() as local_cr:
             pid_col = 'pid' if local_cr.connection.server_version >= 90200 else 'procpid'
-            query = 'SELECT pg_terminate_backend({}) FROM pg_stat_activity WHERE datname=%s'.format(pid_col)
+            try:
+                query = 'SELECT pg_terminate_backend({}) FROM pg_stat_activity WHERE datname=%s'.format(pid_col)
+            except InsufficientPrivilege:
+                _logger.warning('Cannot terminate backend process for %s (maybe a root psql console is accessing the database).', dbname)
```
Xavier-Do, line 883: We could maybe add a runbot warning? this could be like that for a while.
I'm also worried because it could wait for a while an slow down the loop. Not sure
We could add a mechanism like pull_info_failures to avoid spamming the log and avoid useless retry. Just keeping a list of failed database in memory maybe

Xavier-Do (changes requested):

PR #734, opened 2023-02-13
[FIX] runbot: avoid traceback when forcing a preparing batch

d-fence:
When forcing a new batch on a bundle the _force method does not return a batch leading to a traceback in the force_bundle route.

### DIFF
--- a/runbot/controllers/frontend.py
+++ b/runbot/controllers/frontend.py
@@ -210,9 +210,10 @@ class Runbot(Controller):
     def force_bundle(self, bundle, auto_rebase=False, **_post):
         _logger.info('user %s forcing bundle %s', request.env.user.name, bundle.name)  # user must be able to read bundle
         batch = bundle.sudo()._force()
-        batch._log('Batch forced by %s', request.env.user.name)
-        batch._prepare(auto_rebase)
-        return werkzeug.utils.redirect('/runbot/batch/%s' % batch.id)
+        if batch:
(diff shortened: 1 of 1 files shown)


runbot/controllers/frontend.py
```diff
@@ -210,9 +210,10 @@ def bundle(self, bundle=None, page=1, limit=50, **kwargs):
     def force_bundle(self, bundle, auto_rebase=False, **_post):
         _logger.info('user %s forcing bundle %s', request.env.user.name, bundle.name)  # user must be able to read bundle
         batch = bundle.sudo()._force()
-        batch._log('Batch forced by %s', request.env.user.name)
-        batch._prepare(auto_rebase)
-        return werkzeug.utils.redirect('/runbot/batch/%s' % batch.id)
+        if batch:
+            batch._log('Batch forced by %s', request.env.user.name)
+            batch._prepare(auto_rebase)
+            return werkzeug.utils.redirect('/runbot/batch/%s' % batch.id)
```
Xavier-Do, line 216: I think that we should actually return `self.last_batch` in the `_force`method.
in this case this will redirect to the existing batch, but also prepare it immediately without waiting the full time. 

In this case, we will also need to check for the state in _prepare 
```
if not self.state == "preparing":
    return
```

Or if you prefer to avoid to change this behaviour, we can also return the redirect in all case but fallbacking on `last_batch` if `batch` is None

PR #738, opened 2023-03-09
[FIX] runbot: remove useless pip warning

d-fence:
When pip installs a package that contains a script, it goes in `~/.local/bin` directory of the current user. This leads to a log warning to warn the user that the script won't be accessible by other users.

e.g.:
`WARNING: The script pygmentize is installed in '/home/odoo/.local/bin' which is not on PATH.`

This is particularly annoying for documentation builds because it may confuse the users.

### DIFF
--- a/runbot/models/host.py
+++ b/runbot/models/host.py
@@ -126,6 +126,7 @@ class Host(models.Model):
             && mkdir /home/{user} \\
             && chown -R {user}:{user} /home/{user}
             USER {user}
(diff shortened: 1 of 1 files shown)


AntoineVDV:
:+1:

d-fence:
We may go for another solution that is to put the . local/bin in PATH instead.
Because we don't want the wrong script to be called.

PR #767, opened 2023-05-19
[FIX] runbot: download over https

mart-e:
Firefox blocks downloads from http link if you are on an https page

runbot

### DIFF
--- a/runbot/data/runbot_data.xml
+++ b/runbot/data/runbot_data.xml
@@ -47,6 +47,11 @@ admin_passwd=running_master_password</field>
         <field name="value">^((master)|(saas-)?\d+\.\d+)$</field>
     </record>
 
+    <record model="ir.config_parameter" id="runbot.use_ssl">
+        <field name="key">runbot.use_ssl</field>
+        <field name="value">True</field>
+    </record>
+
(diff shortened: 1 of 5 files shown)


Xavier-Do:
The issue is that you are hardcoding https. This is not always correct since some people may have runbot deployed without https.

I started a fix here `https://github.com/odoo/runbot/pull/709/files` but did not get the time to finish it yet since I use this opportunity to migrate to a host_id instead of host on build. I guess that for now we could just use a global icp and add that to the runbot settings instead of a per worker config.

mart-e:
> This is not always correct since some people may have runbot deployed without https.

Deploying over HTTP is having a bad press these days but fair enough.
Or we could set the URL to `//runbot.odoo.com/runbot/...` (I just updated the PR)

Xavier-Do:
> > This is not always correct since some people may have runbot deployed without https.
> 
> Deploying over HTTP is having a bad press these days but fair enough. Or we could set the URL to `//runbot.odoo.com/runbot/...` (I just updated the PR)

This won't work because the log http url is also used to download a database using wget 
https://github.com/odoo/runbot/blob/15.0/runbot/models/build_config.py#L751

As I said I have another pr to solve this issue would be better to continue on this one.

PR #783, opened 2023-06-23
[IMP] runbot: diffable tracking in chatter

Xavier-Do:
The python steps can be long and interesting to track but the change is actually hard to see since a block of code is logged instead of the diff.

Also, the white spaces are not preserved since we are note in a `<pre>` block making it hard to read.

This proposes an alternative to display python code tracking values as a diff with options to copy the raw content of the old and new version. (as well as showing unchanged lines or not)

**Before:**
image

**After:**

image

**With context:**

image


d-fence (approved):
Really really really nice and useful

PR #808, opened 2023-09-04
16.0 runbot improve commit page moc

d-fence:
- [FIX]: ensure that commit date is always set
- [IMP]: allow to call the commit route with a hash
- [IMP]: add a backend commit links tree view


runbot/controllers/frontend.py
```diff
@@ -234,8 +234,13 @@ def slot_create_build(self, slot=None, **kwargs):
...
 
-    @route(['/runbot/commit/<model("runbot.commit"):commit>'], website=True, auth='public', type='http', sitemap=False)
-    def commit(self, commit=None, **kwargs):
+    @route([
+        '/runbot/commit/<model("runbot.commit"):commit>',
+        '/runbot/commit/<string(length=40):commit_hash>'
+    ], website=True, auth='public', type='http', sitemap=False)
+    def commit(self, commit=None, commit_hash=None, **kwargs):
+        if commit_hash:
+            commit = request.env['runbot.commit'].search([('name', '=', commit_hash)])
```
Xavier-Do, line 243: Verry nice!

To give a suggestion, we could redirect to the canonical url (with the id) (using slug) 

Also, could be great to have a `('name', '=like', f'{commit_hash}%')` without the limit if it works t be able to give partial hash.
We may check the length of the result and raise an error if there is more than one

runbot/models/commit.py
```diff
@@ -33,6 +33,13 @@ class Commit(models.Model):
     dname = fields.Char('Display name', compute='_compute_dname')
     rebase_on_id = fields.Many2one('runbot.commit', 'Rebase on commit')
 
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            if 'date' not in vals:
+                vals['date'] = fields.Datetime.now()
```
Xavier-Do, line 40: Could be interesting to get the real info
looks like it can only happen in the 
batch._prepare
batch._update_commits_infos
it the merge_base_sha returns a unknown commit.

d-fence: I kept that one as a safeguard and added a _get_commit_infos method to get the real info

PR #818, opened 2023-10-10
[FIX] runbot: fix inconsistent update global_state

Xavier-Do:
In some case the global state can be inconsistent, waiting when all builds are done. My suspicion is that two build finish at the same time, but the update is not done in database leading to an inconsistent state if two build finishes at the same time.

This should solve this issue by forcing the update, leading to the rollback of one of the transaction.

This still need further investigation to check if it is the real issue.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -261,6 +261,7 @@ class BuildResult(models.Model):
                     record.global_state = record.local_state
                 else:
                     record.global_state = 'waiting'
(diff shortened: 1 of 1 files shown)


Xavier-Do:
After discussion the fact that the query is not executed is desirable,mainly if many child finishes at the same time.

Waiting for a proper queue, another solution could be to add a cron with an add to recompute for all waiting build, every 5 minutes or so.

Xavier-Do:
A cron was added, latter, should be a good idea to use a message queue

PR #852, opened 2024-02-20
[IMP] runbot: reinforce version

d-fence:
When updating a bundle and changing the is_base field to True, it creates a new version based on the bundle name. This can potentially breaks builds and moreover, it can breaks the whole runbot when a duplicate version name is created.

With this commit:
- a constraint on the version name uniqueness is added
- the is_base field is hidden in the interface when the bundle name does not match the regular expression defined in the settings
- a version cannot be created by the compute version if the bundle name does not match the above mentioned regular expression

### DIFF
--- a/runbot/models/version.py
+++ b/runbot/models/version.py
@@ -1,7 +1,7 @@
 import logging
 import re
 from odoo import models, fields, api, tools
-
+from odoo.exceptions import UserError
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


runbot/views/bundle_views.xml
```diff
@@ -50,7 +50,8 @@
                             <field name="project_id"/>
                             <field name="sticky" readonly="0"/>
                             <field name="to_upgrade" readonly="0"/>
-                            <field name="is_base"/>
+                            <field name="match_base" invisible="1"/>
+                            <field name="is_base" invisible="not match_base"/>
```
Xavier-Do, line 54: I would put it visible but readonly in all cases
This is a useful visual information but it makes no sense to modify it most of the time.

runbot/models/version.py
```diff
@@ -25,6 +25,10 @@ class Version(models.Model):
 
     dockerfile_id = fields.Many2one('runbot.dockerfile', default=lambda self: self.env.ref('runbot.docker_default', raise_if_not_found=False))
 
+    _sql_constraints = [
```
Xavier-Do, line 28: we should actually ensure number uniqueness too

PR #904, opened 2024-06-28
[FIX] runbot: add missing opcode

d-fence:
The MAKE_CELL opcode appeared in python 3.11 and is needed in some python steps when using closures and generators.

Like:
`(all(s > e for e in [1,2]) for s in [0,1])`


d-fence:
@odony @Xavier-Do should'nt we add this opcode in Odoo ? 
And #905

PR #907, opened 2024-07-08
[FIX] runbot: patch unidiff 0.7.3

Xavier-Do:
There is an issue in unidiff 0.7.3 fixed in 0.7.4
https://github.com/matiasb/python-unidiff/commit/a3faffc54e5aacaee3ded4565c534482d5cc3465

This version is not able to parse a diff with removed files.

Since the unidiff packaged version in noble is 0.7.3 patching it looks like the easiest solution

### DIFF
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -19,6 +19,18 @@ from odoo.tools.safe_eval import safe_eval, test_python_expr, _SAFE_OPCODES, to_
 # to simplify some python step by wraping the content in a function to allow return statement and get closer to other
 # steps
 
+
+# There is an issue in unidiff 0.7.3 fixed in 0.7.4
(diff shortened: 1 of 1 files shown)


runbot/models/build_config.py
```diff
@@ -19,6 +19,18 @@
...
 # steps
 
+
+# There is an issue in unidiff 0.7.3 fixed in 0.7.4
+# https://github.com/matiasb/python-unidiff/commit/a3faffc54e5aacaee3ded4565c534482d5cc3465
+# Since the unidiff packaged version in noble is 0.7.3
+# patching it looks like the easiest solution
+
+from unidiff import patch, VERSION
+if VERSION == '0.7.3':
```
d-fence, line 29:
```suggestion
from importlib import metadata
from unidiff import patch
if metadata.version('unidiff') == '0.7.3':
```

d-fence (changes requested):
I think that it's safer to use metadata

PR #913, opened 2024-07-18
[WIP] store docker error message in a dedicated model, has well has metadata about result

Xavier-Do:
## Store docker error message in a dedicated model,

The docker failure message is not readable and takes a lot of place in the chatter, lets move it in a dedicated model.

This may also be useful later to have a list of existing docker images identifier, make stats about build time, ...

image

## Rework dockerfile generation

The first docker images composition was based on view to be able to customize the generation logic, but this is kind of unpractical to play with docker files, view, and qweb format for that
    
Mainly, the need to escape && in layers, readability, impact of changes
    
This is a proposal for a new model, meeting the specific needs of dockerfile generations while still covering some of the previous inheritance and rendering possibilities.

image

We still want to keep some compositional, but reducing the complexity by giving a well defined structure

Each line in this view is a layer (even if a layer can contain multiple real docker layer..) 

A layer can be:

### raw
Just use the content
### debian_packages
Give a list of debian packages, the default layer with apt update and cleanup will be used
### pip
Install pip packges with default flags
### template
A little like raw, but with a dictionary of values to fill in the content. Mainly useful to allow to reuse an existing layer but tweaking some values like versions
### reference_layer
Use an existing layer. Can also give some values that will be use instead of the default one if the layer is template
### reference_file
Similar to reference_layer but can link a complete docker files, all layers will be used, with the custom values. 

A reference count is also displayed to know if a layer is referenced anywhere. Will be displayed in yellow if referenced trough the dockerfile, red if imediatly referenced by another layer. This should prevent modifying a layer that is common to multiple dockerfile without paying attention.  The list of reference is available in the view form.

image


Xavier-Do:
Not tested yet, just a poc

PR #921, opened 2024-07-30
[IMP] runbot: allow docker user creation at any place

d-fence:
The docker user is created at the very end of the Dockerfile but, sometimes, at the Docker image build, we need to do stuffs with the regular user instead of root.

With this commit, when the `USER TEMPLATE` string is encountered alone on a line, it's replaced by the user creation lines. The other instances of that string are replaced by the simple USER directive, allowing to easyly switch users during the Docker image build.

Also, the shell user home directory creation is simplified by using the `--create-home` argument.

### DIFF
--- a/runbot/models/host.py
+++ b/runbot/models/host.py
@@ -1,5 +1,6 @@
 import logging
 import getpass
+import re
 import time
 
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


Xavier-Do:
Will be done in #913

The idea is that the user creation will be a layer making it less obscure

PR #940, opened 2024-09-02
WIP - add the possibility to disable a test per version

d-fence:
When disabling tests on runbot by using the test_tags on a build error, the tests are disabled on every tested version. This can be annoying when a test only fails from one version up to another.

With this commit, a min and max version can be specified on a build_error when the test_tags field is used.
If the min and max are not used, the test will be disabled cross versions.

### DIFF
--- a/runbot/models/build_error.py
+++ b/runbot/models/build_error.py
@@ -73,12 +73,20 @@ class BuildError(models.Model):
     last_seen_build_id = fields.Many2one('runbot.build', compute='_compute_last_seen_build_id', string='Last Seen build', store=True)
     last_seen_date = fields.Datetime(string='Last Seen Date', compute='_compute_seen_date', store=True)
     test_tags = fields.Char(string='Test tags', help="Comma separated list of test_tags to use to reproduce/remove this error", tracking=True)
+    tags_min_version_id = fields.Many2one('runbot.version', 'Tags Min version', help="Minimal version where the test tags will be applied.")
+    tags_max_version_id = fields.Many2one('runbot.version', 'Tags Max version', help="Maximal version where the test tags will be applied.")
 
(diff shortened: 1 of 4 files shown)


runbot/models/build_error.py
```diff
@@ -263,14 +272,21 @@ def _parse_logs(self, ir_logs):
...
-    def _test_tags_list(self):
-        active_errors = self.search([('test_tags', '!=', False)])
+    def _test_tags_list(self, build_id=False):
+        tags_domain = [('test_tags', '!=', False)]
+        if build_id:
+            tags_domain = expression.AND([tags_domain, [('tags_min_version_id', '=', False), ('tags_max_version_id', '=', False)]])
+            version_id = build_id.params_id.version_id
+            tags_domain = expression.OR([tags_domain, [('test_tags', '!=', False), ('tags_max_version_id.number', '>=', version_id.number), ('tags_min_version_id.number', '<=', version_id.number)]])
+            tags_domain = expression.OR([tags_domain, [('test_tags', '!=', False), ('tags_max_version_id', '=', False), ('tags_min_version_id.number', '<=', version_id.number)]])
+            tags_domain = expression.OR([tags_domain, [('test_tags', '!=', False), ('tags_max_version_id.number', '>=', version_id.number), ('tags_min_version_id', '=', False)]])
```
```suggestion
        tags_domain = [('test_tags', '!=', False)]
        if build_id:
            version_id = build_id.params_id.version_id
            tags_domain = expression.AND([tags_domain, [
            '|', 
                ('tags_min_version_id', '=', False),
                ('tags_min_version_id.number', '<=', version_id.number),
            '|', 
                ('tags_max_version_id', '=', False),
                ('tags_max_version_id.number', '>=', version_id.number),
            ]])
```

PR #958, opened 2024-09-26
[IMP] runbot: display version in build error view

davidmonnom:
Before this commit, the build error view was not displaying the version of versions that was buggy. This commit adds the version in the build error view.

### DIFF
--- a/runbot/templates/build_error.xml
+++ b/runbot/templates/build_error.xml
@@ -9,6 +9,7 @@
               <div class="col">Last seen date</div>
               <div class="col col-md-3">Module</div>
               <div class="col col-md-5">Summary</div>
+              <div class="col">Version</div>
               <div class="col">Triggers</div>
               <div class="col">Assigned to</div>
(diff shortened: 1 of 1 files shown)


runbot/templates/build_error.xml
```diff
@@ -26,6 +27,11 @@
                     <code><t t-esc="build_error.summary"/></code>
                   </button>
                 </div>
+                <div class="col">
+                  <t t-foreach="build_error.version_ids" t-as="version">
+                    <span t-attf-class="{{version.is_major ? 'text-bg-warning' : 'text-bg-info'}}" class="badge text-bg-pill small"><t t-esc="version.name"/></span>
```
d-fence:
```suggestion
                    <span t-attf-class="{{'text-bg-warning' if version.is_major else 'text-bg-info'}}" class="badge text-bg-pill small"><t t-esc="version.name"/></span>
```

d-fence (changes requested):

PR #967, opened 2024-10-22
[IMP] runbot: refactor build error models

Xavier-Do:
The initial idea to link an error to another one was a quick solution to group them if they where related, but this became challenging to compute metada regarding errors.

- The displayed error message was not always consistent with the real root cause/the error that lead here.
- The aggregates (lets says, linked buils ids) could be the one of the error, or from all error messages. Same for the versions, first seen, ..
This is confusing to knwo what is the leist we are managing and what is the expecte result to display

Main motivation:
on a standard error page (will be changed to "assignment"), we want to have the list of error message that is related to this one. We want to know for each message (a real build error) what is the version, first seen, ...
This will give more flexibility on the display,

The assigned person/team/test-tags, ... are moved to this model
The appearance data remains on the build error but are aggregate on the assignation.


runbot/migrations/17.0.5.8/post-migration.py
```diff
@@ -0,0 +1,123 @@
...
+        children = error_by_parent.get(error_id, [])
+        assert not error_id in children
+        all_errors = [error_id, *children]
+        error_count = len(all_errors)
+
+        # vals_by_error order: min(build_id), min(log_date), max(build_id), max(log_date)
+        build_count = 0
+        first_seen_build_id = first_seen_date = last_seen_build_id = last_seen_date = None
+        if error_id in vals_by_error:
+            error_vals = [vals_by_error[error_id] for error_id in all_errors]
```
d-fence:
```suggestion
            error_vals = [vals_by_error[error_id] for error_id in all_errors if error_id in vals_by_error]
```
To avoid script failure when to the error has no build linked.

d-fence (changes requested):

PR #969, opened 2024-10-25
[REF] runbot: remove link type ir.logging

Williambraecky:
1: Stop using the special \$\$icon\$ \$ syntax when logging the download url for
the database dump.

2: Removes the link type in ir.logging as they were using some weird dark
magic instead of the more generic markdown logging type.

### DIFF
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -925,13 +925,15 @@ class ConfigStep(models.Model):
             build._logger('Step %s finished in %s' % (self.name, s2human(build.job_time)))
             return
 
-        kwargs = dict(message='Step %s finished in %s' % (self.name, s2human(build.job_time)))
+        message = 'Step %s finished in %s'
+        args = [self.name, s2human(build.job_time)]
+        log_type = 'runbot'
         if self.job_type == 'install_odoo':
-            kwargs['message'] += ' $$fa-download$$'
             db_suffix = build.params_id.config_data.get('db_name') or (build.params_id.dump_db.db_suffix if not self.create_db else False) or self.db_name
-            kwargs['path'] = '%s%s-%s.zip' % (build._http_log_url(), build.dest, db_suffix)
-            kwargs['log_type'] = 'link'
-        build._log('', **kwargs)
+            message += ' [@icon-download](%s%s-%s.zip)'
+            args += [build._http_log_url(), build.dest, db_suffix]
+            log_type = 'markdown'
+        build._log('', message, *args, log_type=log_type)
 
         if self.coverage:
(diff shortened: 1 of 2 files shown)


Williambraecky:
@Xavier-Do Came across this while reading through the runbot repository and found it could use a little bit of cleanup.

I made the migration in a separate commit in case we would rather not do the migration.

runbot/migrations/17.0.5.8/pre-migrate.py
```diff
@@ -0,0 +1,32 @@
...
+                        message,
+                        '(.*)\$\$fa-([a-zA-Z]*)\$\$(.*)',
+                        CONCAT('\1[@icon-\2](', "path", ')\3')
+                    ),
+                    '$$',
+                    ''
+                ),
+               type = 'markdown'
+         WHERE type = 'link'
+    """, 'ir_logging', bucket_size=100000)
```
Xavier-Do: We don't use utils for upgrades on runbot usually I would prefer to avoid it ;) Anyway the ir_logging table is hudge so I'm not sure we need to migrate it completely. We can test it if not done yet. This can be done after the upgrade asynchronously.

runbot/models/build_config.py
```diff
@@ -884,10 +884,9 @@ def _log_end(self, build):
 
         kwargs = dict(message='Step %s finished in %s' % (self.name, s2human(build.job_time)))
         if self.job_type == 'install_odoo':
-            kwargs['message'] += ' $$fa-download$$'
             db_suffix = build.params_id.config_data.get('db_name') or (build.params_id.dump_db.db_suffix if not self.create_db else False) or self.db_name
-            kwargs['path'] = '%s%s-%s.zip' % (build._http_log_url(), build.dest, db_suffix)
-            kwargs['log_type'] = 'link'
+            kwargs['message'] += ' [@icon-download](%s)' % ('%s%s-%s.zip' % (build._http_log_url(), build.dest, db_suffix))
+            kwargs['log_type'] = 'markdown'
         build._log('', **kwargs)
```
Xavier-Do: In the case of markdown logs, _http_log_url, dest, suffix should ideally be passed as arguments to the  build._log method so that it is escaped. This will avoid to potentially escape parts of the _http_log_url

In this case it is not a big issue I think but it would be cleaner

```suggestion
        build._log('',  ' [@icon-download](%s)' % ('%s%s-%s.zip' , build._http_log_url(), build.dest, db_suffix, log_type= 'markdown')
```

Xavier-Do (changes requested):
Nice cleanup, it was time to remove this old hack indeed! 

Small comments but we can check that together Wednesday and test everything.

PR #970, opened 2024-10-25
[IMP] runbot: permit to qualify build errors

d-fence:
Identifying build errors by fingerprint is not enough. Most of the times we want to link build errors based on some criteria.

With this commit, a build error can be qualified by using regular expressions that will extract informations.

Those informations will be stored in a jsondict field. That way, we can find simmilar build errors when they share some qualifiers.

To extract information, the regular expression must use named group patterns.

e.g:
`^Tour (?P<tour_name>\w+) failed at step (?P<tour_step>.+)`

The above regular expression should extract the tour name and tour step from a build error and store them like:
`{ "tour_name": "article_portal_tour", "tour_step": "clik on 'Help'" }`

The field can be queried to find similar errors. Also, a button is added on the Build Error Form to search for errors that share the same qualifiers.

**TODO**
* Decide if we can select the source field or if it will be always based on the content field
* Decide if we store the regex_id alongside with the qualifier (probably not a good idea)


d-fence:
Some examples of regexes
```
^Traceback \(most recent call last\):\n  File \"/data/build/(enterprise|odoo/addons)/(?P<module>\w+)/tests/(?P<filename>\w+\.py)\",
^(Error received after termination: )?Tour (?P<tour_name>\w+) failed at step (?P<tour_step>.+)
AssertionError: The test code \"odoo\.startTour\('(?P<tour_name>\w+)'
^(ERROR|FAIL): (?P<test_class>\w+)\.(?P<test_method>\w+)
^(ERROR|FAIL): Subtest (?P<test_class>\w+)\.(?P<test_method>\w+) (?P<subtest_params>.+)\n
ERROR: setUpClass \(odoo\.addons\.(?P<module>\w+)\.tests\..+\.(?P<test_class>\w+)\)
module (?P<module>\w+): not installable.+
^Module (?P<module>\w+) demo data failed to install.+
FAILED: \[\d+/\d+\] Tour (?P<tour_name>\w+).+Step (?P<tour_step>.+)( |\.)(-+|\(|\n)
\[HOOT\] Test \"(?P<hoot_test>.+)\" failed
```

runbot/models/build_error.py
```diff
@@ -618,3 +707,76 @@ def action_submit(self):
...
+    def _compute_qualifiers(self):
+        for record in self:
+            if record.regex and record.test_string:
+                record.qualifiers = record._qualify(record.test_string)
+            else:
+                record.qualifiers = {}
+
+
+class QualifyErrorSampleLink(models.Model):
+    _name = 'runbot.error.sample.link'
```
Xavier-Do: `"runbot.error.qualify.test"` ?

d-fence: Yes, it seems better. I change to that

PR #988, opened 2024-11-19
[IMP] runbot: get docker metadata after build

Xavier-Do:
The docker metadata are currently computed only on **some** images during the nightly.

This pr aims to get metadata after each docker image build in order to be able to rely on them when needed, having metadata on **all** images **immediately**.

### DIFF
--- a/runbot/models/docker.py
+++ b/runbot/models/docker.py
@@ -2,6 +2,7 @@ import getpass
 import logging
 import os
 import re
+import docker
 from odoo import api, fields, models
 from odoo.addons.base.models.ir_qweb import QWebException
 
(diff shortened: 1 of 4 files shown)


runbot/models/docker.py
```diff
@@ -155,8 +156,25 @@ def copy(self, default=None):
...
+                    max(id)
+                FROM
+                    runbot_docker_build_result
+                WHERE
+                    result = 'success'
+                    AND dockerfile_id IN %s
+                GROUP BY dockerfile_id
+                """, [tuple(self.ids)]
+            )
+            results = self.env['runbot.docker_build_result'].browse([r[0] for r in self.env.cr.fetchall()])
```
Williambraecky: Should be good
```suggestion
            rg = self.env['runbot.docker_build_result']._read_group(
                [('result', '=', 'success'), ('dockerfile_id', 'in', self.ids)],
                ['dockerfile_id'],
                ['id:max'],
            )
            result_ids = self.env['runbot.docker_build_result'].browse([r[1] for r in rg])
```

Williambraecky (approved):

PR #996, opened 2024-11-27
[IMP] runbot: make computable fields searchable

Williambraecky:
Adds search functions for all 'related' computed fields on runbot.build.error.


d-fence (approved):
lgtm

PR #1001, opened 2024-11-29
[IMP] runbot: add without-demo as field

Williambraecky:
Adds the without-demo flag as a field on config steps.

### DIFF
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -160,6 +160,10 @@ class ConfigStep(models.Model):
     extra_params = fields.Char('Extra cmd args', tracking=True)
     additionnal_env = fields.Char('Extra env', help='Example: foo="bar";bar="foo". Cannot contains \' ', tracking=True)
     enable_log_db = fields.Boolean("Enable log db", default=True)
+    demo_mode = fields.Selection(
+        [('default', 'Default'), ('without_demo', 'Without Demo'), ('with_demo', 'With Demo')],
+        "Install demo data", default='default', tracking=True, required=True,
+    )
     # python
     python_code = fields.Text('Python code', tracking=True, default=PYTHON_DEFAULT)
     python_result_code = fields.Text('Python code for result', tracking=True, default=PYTHON_DEFAULT)
@@ -431,13 +435,23 @@ class ConfigStep(models.Model):
         if self.create_db:
             build._local_pg_createdb(db_name)
         cmd += ['-d', db_name]
+
+        # Demo data behavior changed in 18.1 -> demo data became opt-in instead of opt-out
+        available_options = build._parse_config()
+        # True if build has demo data by default
+        demo_installed_by_default = '--with-demo' not in available_options
+        demo_mode = build.params_id.config_data.get('demo_mode', self.demo_mode)
+        if demo_mode == 'with_demo' and not demo_installed_by_default:
+            cmd.append('--with-demo')
+        elif demo_mode == 'without_demo' and demo_installed_by_default:
+            cmd.append('--without-demo=true')
+
         # list module to install
         extra_params = build.params_id.extra_params or self.extra_params or ''
         if mods and '-i' not in extra_params:
             cmd += ['-i', mods]
         config_path = build._server("tools/config.py")
 
-        available_options = build._parse_config()
         if self.test_enable:
             if "--test-enable" in available_options:
                 cmd.extend(['--test-enable'])
--- a/runbot/models/batch.py
+++ b/runbot/models/batch.py
@@ -372,7 +372,7 @@ class Batch(models.Model):
             # in any case, search for an existing build
             config = trigger_custom.config_id or trigger.config_id
             extra_params = trigger_custom.extra_params or ''
-            config_data = trigger_custom.config_data or {}
+            config_data = dict(trigger.config_data or {}) | dict(trigger_custom.config_data or {})
             params_value = {
                 'version_id':  version_id,
                 'extra_params': extra_params,
--- a/runbot/models/repo.py
+++ b/runbot/models/repo.py
@@ -14,6 +14,7 @@ from pathlib import Path
 from odoo import models, fields, api
 from odoo.tools import file_open, mail
 from ..common import os, RunbotException, make_github_session, sanitize
+from ..fields import JsonDictField
 from odoo.exceptions import UserError
 from odoo.tools.safe_eval import safe_eval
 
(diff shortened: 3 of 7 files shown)


runbot/models/build_config.py
```diff
@@ -431,6 +432,11 @@ def _run_install_odoo(self, build):
         if self.create_db:
             build._local_pg_createdb(db_name)
         cmd += ['-d', db_name]
+        # Demo data behavior changed in 18.1 -> demo data became opt-in instead of opt-out
+        if build.version_id.number >= '18.01' and self.with_demo:
```
Williambraecky: This will not work obviously with the PR but I have no idea how we can make it work "temporarily"

runbot/models/build_config.py
```diff
@@ -160,6 +160,7 @@ class ConfigStep(models.Model):
     extra_params = fields.Char('Extra cmd args', tracking=True)
     additionnal_env = fields.Char('Extra env', help='Example: foo="bar";bar="foo". Cannot contains \' ', tracking=True)
     enable_log_db = fields.Boolean("Enable log db", default=True)
+    with_demo = fields.Boolean("Install demo data", default=True, tracking=True)
```
Xavier-Do: I'm not sure that having a default True is correct since we may want to have the default behaviour most of the time. (single module, testing, install of base database, at install, ....)

We may want to be able to force without in previous version, in this case the flag --without is already added 
We may want to force with demo on future version, but the only use case would be to have a database for the run (and maybe a nightly) and this will be a contionnal step. (have a database with demo for the running build) 

But adding it on the config could be a nice idea, I think that in this case we should also add support for `build.params_id.config_params.get('no_demo')`. But the field may be a selection field (default, no_demo, demo) so that the odoo testing can run without demo. Upgrade template will force "no demo" and master default config could install a template with demo on the side for the running build.

runbot/models/build_config.py
```diff
@@ -768,6 +774,11 @@ def _run_test_upgrade(self, build):
         upgrade_paths = list(build._get_upgrade_path())
         if upgrade_paths:
             migrate_cmd += ['--upgrade-path', ','.join(upgrade_paths)]
+        # Demo data behavior changed in 18.1 -> demo data became opt-in instead of opt-out
+        if build.version_id.number >= '18.01' and self.with_demo:
+            migrate_cmd.append('--with-demo')
+        elif build.version_id.number < '18.01' and not self.with_demo:
+            migrate_cmd.append('--without-demo=true')
```
Xavier-Do:
```suggestion
        # Demo data behavior changed in 18.1 -> demo data became opt-in instead of opt-out
        available_options = build._parse_config()
        if self.with_demo and '--with-demo' in available_options:
            migrate_cmd.append('--with-demo')
        elif not self.with_demo and '--with-demo' not in available_options:
            migrate_cmd.append('--without-demo=true')
```

runbot/models/build_config.py
```diff
@@ -431,6 +432,11 @@ def _run_install_odoo(self, build):
         if self.create_db:
             build._local_pg_createdb(db_name)
         cmd += ['-d', db_name]
+        # Demo data behavior changed in 18.1 -> demo data became opt-in instead of opt-out
+        if build.version_id.number >= '18.01' and self.with_demo:
+            cmd.append('--with-demo')
+        elif build.version_id.number < '18.01' and not self.with_demo:
+            cmd.append('--without-demo=true')
```
Xavier-Do:
```suggestion
        available_options = build._parse_config()
        if self.with_demo and '--with-demo' in available_options:
            migrate_cmd.append('--with-demo')
        elif not self.with_demo and --with-demo' not in available_options:
            migrate_cmd.append('--without-demo=true')
```

Xavier-Do (changes requested):

runbot/models/build_config.py
Xavier-Do: Another solution could be to have two boolean fields
without-demo and with demo, both of them being False by default.

Williambraecky: Thanks for the insight 😄 

I like the selection field as it provides the most flexibility so I implemented it and tested all different ways.
Let me know what you think :)

Xavier-Do: The selection field doesn't allow an incorrect state where the two are True but it could have a 4th effect by design, do the opposite (no demo if demo is default, and vice versa)

Williambraecky: Do you want me to add that 4th state?

Xavier-Do: Not sure. Would be confusing. We can talk about that wednesday

runbot/models/build_config.py
```diff
@@ -431,13 +435,24 @@ def _run_install_odoo(self, build):
...
+        available_options = build._parse_config()
+        build_default_demo_behavior = 'without_demo' if '--with-demo' in available_options else 'with_demo'
+        demo_mode = build.params_id.config_data.get('demo_mode', self.demo_mode)
+        wanted_demo_behavior = build_default_demo_behavior if demo_mode == 'default' else demo_mode
+        if wanted_demo_behavior != build_default_demo_behavior:
+            if wanted_demo_behavior == 'with_demo':
+                cmd.append('--with-demo')
+            else:
+                cmd.append('--without-demo=true')
+
```
Xavier-Do:
```suggestion
        available_options = build._parse_config()
        demo_with_is_default = '--with-demo' not in available_options:
        demo_mode = build.params_id.config_data.get('demo_mode', self.demo_mode)
        if wanted_demo_behavior != build_default_demo_behavior:
            if demo_mode == 'with_demo' and not demo_with_is_default:
                cmd.append('--with-demo')
            elif demo_mode == 'without_demo'  and demo_with_is_default
                cmd.append('--without-demo=true')
```

Not sure, maybe?

runbot/models/build_config.py
```diff
@@ -160,6 +160,10 @@ class ConfigStep(models.Model):
     extra_params = fields.Char('Extra cmd args', tracking=True)
     additionnal_env = fields.Char('Extra env', help='Example: foo="bar";bar="foo". Cannot contains \' ', tracking=True)
     enable_log_db = fields.Boolean("Enable log db", default=True)
+    demo_mode = fields.Selection(
+        [('default', 'Default'), ('without_demo', 'Without Demo'), ('with_demo', 'With Demo')],
+        "Install demo data", default='default', tracking=True
```
Xavier-Do: `required=True`

PR #1007, opened 2024-12-05
[IMP] runbot: qualify errors on selected field

d-fence:
The previous qualifying implementation was only qualifying errors based on error content. With this commit the field on which the qualifying applies can be selected on a qualifying regex.


Williambraecky (approved):
LGTM 👍 
`summary` in build.error.content uses 80 characters while you are using 70 but it doesn't really matter much.

d-fence:
Finally another idea emerged by discussing with xdo.
The user should be able to choose multiple fields on which the regex apply.
The use case is to avoid having a too "weak" qualifier. e.g.: a wrong module is worst that no module at all.

PR #1008, opened 2024-12-05
[PERF] runbot: remove row_number queries

Williambraecky:
The row_number window function is quite unoptimized as it requires reading the whole table.
Using lateral join and / or distinct makes better use of existing indexes.

### DIFF
--- a/runbot/models/bundle.py
+++ b/runbot/models/bundle.py
@@ -144,65 +144,55 @@ class Bundle(models.Model):
 
     @api.depends_context('category_id')
     def _compute_last_batchs(self):
-        batch_ids = defaultdict(list)
+        batches_by_bundle = defaultdict(lambda: self.env['runbot.batch'])
         if self.ids:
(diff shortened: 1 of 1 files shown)


Williambraecky:
@Xavier-Do would be worth it to run an explain (analyse) on those queries on prod/mirror just to check if it saves any time at all.

runbot/models/bundle.py
```diff
@@ -144,65 +144,49 @@ def _compute_relations_base_id(self):
...
-                WHERE
-                    row <= 4
-                ORDER BY row, id desc
-                """, [tuple(self.ids), category_id]
-            )
-            batchs = self.env['runbot.batch'].browse([r[0] for r in self.env.cr.fetchall()])
-            for batch in batchs:
-                batch_ids[batch.bundle_id.id].append(batch.id)
-
+                SELECT bundle.id, ARRAY_AGG(batch.id)
```
Xavier-Do, line 151: Similar in perfs to the previous, except with  `create index runbot_batch__bundle_id_id_desc_index on runbot_batch (bundle_id, id desc);` Blazing fast !!!!!!

PR #1009, opened 2024-12-06
[IMP] runbot: add a widget for pull requests url's

d-fence:
While it's useful to have a clickable link to the fixing pull request on the runbot error list, we cannot display the entire url as it takes too much space. So a static text `View PR` is used.

As this static text is of limited use, this commit adds a new widget that displays a shortened URL for the github pull request only showing the repo and the PR number.

### DIFF
--- a/runbot/static/src/js/fields/fields.js
+++ b/runbot/static/src/js/fields/fields.js
@@ -14,6 +14,7 @@ import { useRef, xml, Component } from "@odoo/owl";
 import { useAutoresize } from "@web/core/utils/autoresize";
 import { getFormattedValue } from "@web/views/utils";
 
+import { UrlField } from "@web/views/fields/url/url_field";
 
 function stringify(obj) {
     return JSON.stringify(obj, null, '\t')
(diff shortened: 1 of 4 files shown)


runbot/static/src/js/fields/fields.js
```diff
@@ -135,6 +136,29 @@ registry.category("fields").add("char_frontend_url", {
...
+    static template = xml`
+    <a class="o_field_widget o_form_uri" t-on-click.stop="" t-att-href="formattedHref" t-esc="props.record.data[props.name].replace('https://github.com/odoo','').replace('/pull','') || ''" target="_blank"/>
+    `;
+    static components = { UrlField }
+    setup() {
+        if (!this.props.readonly) {
+            throw new Error("This widget works only on readonly fields");
+        }
+    }
+}
```
Williambraecky: I would consider this instead:
 - Use a regex to get the parts of the url we need.
 - Use the same format as used on github <user>/<repo>#issueId.
 - Remove custom template -> prefer using the existing field and compute the `text` props ourselves.

I removed the readonly constraint but we could definitely keep it.

```suggestion
// Pull Request URL Widget
const pullRequestRegex = /\/([a-zA-Z-_]+\/[a-zA-Z-_]+)\/pull\/(\d+)/;
class PullRequestUrlField extends UrlField {
    static template = xml`
        <UrlField t-props="fieldProps"/>
    `;
    static components = { UrlField }

    get fieldProps() {
        const props = {...this.props};
        const parts = pullRequestRegex.exec(this.props.record.data[props.name])
        if (parts) {
            props.text = `${parts[1]}#${parts[2]}`;
        }
        return props
    }
}
```

Williambraecky (approved):
LGTM :)

Williambraecky (approved):
@d-fence Since we display branches with yet another format I added support for that format in the branch search name.

PR #1010, opened 2024-12-09
[IMP] runbot: 18.0 frontend improvements

Williambraecky:


Williambraecky:
@Xavier-Do most of those have no real impact on 18.0 and could be merged later.

But I would consider keeping `[FIX] runbot: keep Chart.js v4.3.0` as chart.js was updated to (IIRC) 4.4.3 in between 17.0 and 18.0 and I am not sure everything has been tested.
Note that the stats page is a very good candidate for a OWL rewrite.

Williambraecky:
The last two commits are important for 18.0 deployment.

Williambraecky:
Will make a new branch with all changes

PR #1016, opened 2024-12-12
[FIX] runbot: fix branch search

Williambraecky:
Allows searching branch through pr url or full branch name. Also fixes a crash when searching invalid values.

### DIFF
--- a/runbot/models/branch.py
+++ b/runbot/models/branch.py
@@ -49,11 +49,19 @@ class Branch(models.Model):
             branch.dname = '%s:%s' % (branch.remote_id.short_name, branch.name)
 
     def _search_dname(self, operator, value):
-        if ':' not in value:
-            return [('name', operator, value)]
-        repo_short_name, branch_name = value.split(':')
-        owner, repo_name = repo_short_name.split('/')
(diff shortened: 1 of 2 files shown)


runbot/models/branch.py
```diff
@@ -49,11 +49,19 @@ def _compute_dname(self):
...
 
     def _search_dname(self, operator, value):
-        if ':' not in value:
-            return [('name', operator, value)]
-        repo_short_name, branch_name = value.split(':')
-        owner, repo_name = repo_short_name.split('/')
-        return ['&', ('remote_id', '=', self.env['runbot.remote'].search([('owner', '=', owner), ('repo_name', '=', repo_name)]).id), ('name', operator, branch_name)]
+        # Match format (owner?, repo, branch)
+        owner = repo = branch = None
+        if (m := re.match(r'(?:([\w-]+)/)?([\w-]+):([\w-]+)', value)):
```
d-fence:
```suggestion
        if (m := re.match(r'(?:([\w-]+)/)?([\w-]+):([\w\.-]+)', value)):
```
Branch names may contains `. ` otherwise, searching for a branch with `odoo:17.0-fix-too` will lead to a `ilike %17%`

Williambraecky: Good catch 😄

Williambraecky:
@d-fence I also added tests to match the new behavior :)

PR #1017, opened 2024-12-12
[FIX] runbot: change md_description to html

Williambraecky:
Otherwise it doesn't output raw html.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -155,7 +155,7 @@ class BuildResult(models.Model):
     # could be a default value, but possible to change it to allow duplicate accros branches
 
     description = fields.Char('Description', help='Informative description')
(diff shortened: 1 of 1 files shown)


Xavier-Do:
Do we need the sanitize = False? isn't the sanitize to remove script and other stuff?

Williambraecky:
> Do we need the sanitize = False? isn't the sanitize to remove script and other stuff?

We don't need it since it is properly handled by our `pseudo_markdown` function.
Without sanitize it will always try to wrap the content with `<p>` which causes layout issues. I did test with injections and nothing bad happens :)
See:
https://github.com/odoo/runbot/blob/18.0-fix-description-md-html-wbr/runbot/tests/test_build.py#L236-L244
https://runbot.odoo.com/runbot/build/71916919

Xavier-Do:
Ok so it is to avoid the `<p>` :) ok for me

PR #1025, opened 2024-12-24
[IMP] runbot: improve min version management

Xavier-Do:
The main goal of this pr is to keep master disabled tests in the freezed version. 

Before this change, during the freeze, any test-tag only enabled on master wouldn't be enable on the freezed version


Williambraecky:
@Xavier-Do I think this PR is ready for review no? Feature is good :)

PR #1031, opened 2025-01-06
[FIX] runbot: fix frontend team dashboards again

d-fence:
The previous fix in c9e8e1360 was not enough, while preventing a crash of the view, the view was empty.


Williambraecky (approved):
Tested locally 👍

PR #1033, opened 2025-01-08
[FIX] runbot: remove unknown compute

Williambraecky:
bundle_id is not computed anymore


d-fence:
Compute was removed in ecfdb56

PR #1034, opened 2025-01-08
[IMP] runbot: add support for attributes on md url

Williambraecky:
Adds support for attributes on our markdown implementation, mainly to add support for target="_blank" on urls.
This can be used to change the behavior of links when we want to differentiate between navigation and information.

### DIFF
--- a/runbot/tests/test_utils.py
+++ b/runbot/tests/test_utils.py
@@ -1,0 +1,49 @@
+from odoo.tests.common import TransactionCase, Like
+
+from odoo.addons.runbot.common import pseudo_markdown
+
+
+class TestUtils(TransactionCase):
+
+    def test_md_formatting(self):
+        self.assertEqual(
+            pseudo_markdown(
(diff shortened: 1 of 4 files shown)


runbot/tests/test_utils.py
```diff
@@ -0,0 +1,55 @@
...
+        # Test with attributes
+        self.assertEqual(
+            pseudo_markdown(
+                "[name](https://runbot.odoo.com){target _blank}"
+            ),
+            '<a href="https://runbot.odoo.com" target="_blank">name</a>'
+        )
+
+        # Invalid input
+        with mute_logger('odoo.addons.runbot.common'):
```
d-fence:
```suggestion
        with self.assertLogs(level='WARNING'):
```
while at it, why not assert that the warning was triggered

Williambraecky:
Not that useful actually, we have no real usecase for it as of now.

PR #1035, opened 2025-01-09
[REF] runbot: rewrite stats page using owl

Williambraecky:
Removing very old javascript and using new tools will allow us to more efficiently change and maintain the stat page in the futur. This commit is meant to produce a 1:1 copy of the stats page like it was before.
Further refactoring will happen to improve performance, layout and component logic.


Williambraecky:
@Xavier-Do This is a big refactor but it does add displaying stats from different bundles at the same time.
I understand owl is not everyone's cup of tea but the previous legacy js code was honestly bad 😅 

I think it's a step in the right direction if we ever want to add more features to that page.

PR #1038, opened 2025-01-20
[REF] runbot: remove deprecated calls

Williambraecky:
Use `_read_group` instead of `read_group` as it is deprecated and will be removed between 18 and 19.


Xavier-Do:
You have conflicts unfortunately :)

Williambraecky:
> You have conflicts unfortunately :)

😩 for a comma
🥷

PR #1045, opened 2025-01-31
[IMP] runbot: remove auto tag if fixing match

Williambraecky:
Removes auto tags from a build if the build's branches match the build error's fixing branch.

### DIFF
--- a/runbot/models/build_error.py
+++ b/runbot/models/build_error.py
@@ -209,8 +209,11 @@ class BuildError(models.Model):
     @api.model
     def _test_tags_list(self, build_id=False):
         version = build_id.params_id.version_id.number if build_id else False
+        branches = build_id.create_batch_id.bundle_id.branch_ids if build_id else self.env['runbot.branch']
 
         def filter_tags(e):
+            if e.fixing_pr_id in branches:
+                return False
             if version:
                 min_v = e.tags_min_version_id.number or ''
                 max_v = e.tags_max_version_id.number or '~'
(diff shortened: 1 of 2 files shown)


runbot/models/build_error.py
```diff
@@ -209,8 +209,11 @@ def _merge(self, others):
     @api.model
     def _test_tags_list(self, build_id=False):
         version = build_id.params_id.version_id.number if build_id else False
+        branches = build_id.params_id.commit_link_ids.branch_id if build_id else self.env['runbot.branch']
```
```suggestion
        branches = build_id.create_batch_id.bundle_id.branch_ids if build_id else self.env['runbot.branch']
```

I wonder if this versions would be better
The reason is that commit_link can be dynamically create in some case, I'm not 100% sure that the branch is always correct. (upgrades mainly) 
A second reason is that if we create an empty branch with an empty batch (commit link corresponding to the base version) a rebuild would be enough after after adding the branch to make it run without the test-tag.

Williambraecky: I wanted to use it at first but actually we don't always populate `create_batch_id` so I am not sure what it is used for:
```
runbot_db=# select count(*), create_batch_id is null from runbot_build group by create_batch_id is null;
  count   | ?column?
----------+----------
 11809290 | f
 43247753 | t
(2 rows)
```

Xavier-Do: I think that most of them are old ones, new ones should always have that.

Actually in the last 6 months we only have a few without create_batch_id, the last one being  https://runbot.odoo.com/runbot/batch/348176/build/70214325
And it is actually a rebuild from another one from 2019... 

But again I'm not sure that it is a better solution than using the commit_link_ids, to discuss

PR #1046, opened 2025-01-31
[IMP] runbot: add qualifiers on build error

d-fence:
As error content's can be qualified, it's now time to use it to manage build errors.

With this commit, a `qualifiers` field is added on the build error and can be used to find other errors that match those qualifiers.

A button to suggest the qualifiers is added. It works by suggesting the common qualifiers among error contents linked to the build error.

Also, once similar error are found, a button can be used to merge those similar errors with the current one.

### DIFF
--- a/runbot/models/build_error.py
+++ b/runbot/models/build_error.py
@@ -103,6 +103,13 @@ class BuildError(models.Model):
     tags_min_version_id = fields.Many2one('runbot.version', 'Tags Min version', help="Minimal version where the test tags will be applied.")
     tags_max_version_id = fields.Many2one('runbot.version', 'Tags Max version', help="Maximal version where the test tags will be applied.")
 
+    common_qualifiers = JsonDictField('Common Qualifiers', compute='_compute_common_qualifiers', store=True, help="Minimal qualifiers in common needed to link error content.")
+    similar_ids = fields.One2many('runbot.build.error', compute='_compute_similar_ids', string="Similar Errors", help="Similar Errors based on common qualifiers")
+    similar_content_ids = fields.One2many('runbot.build.error.content', compute='_compute_similar_content_ids', string="Similar Error Contents", help="Similar Error contents based on common qualifiers")
+    unique_qualifiers = JsonDictField('Non conflicting Qualifiers', compute='_compute_unique_qualifiers', store=True, help="Non conflicting qualifiers in common needed to link error content.")
+    analogous_ids = fields.One2many('runbot.build.error', compute='_compute_analogous_ids', string="Analogous Errors", help="Analogous Errors based on unique qualifiers")
+    analogous_content_ids= fields.One2many('runbot.build.error.content', compute='_compute_analogous_content_ids', string="Analogous Error Contents", help="Analogous Error contents based on unique qualifiers")
+
     # Build error related data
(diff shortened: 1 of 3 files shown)


runbot/models/build_error.py
```diff
@@ -242,6 +259,42 @@ def action_view_errors(self):
...
+            all_qualifiers = [r.qualifiers.dict for r in record.error_content_ids]
+            common_keys = set.intersection(*map(set, [q.keys() for q in all_qualifiers]))
+            if common_keys:
+                infered_qualifiers = dict()
+                for k in common_keys:
+                    values = {q.get(k) for q in all_qualifiers}
+                    if len(values) == 1:
+                        infered_qualifiers[k] = values.pop()
+                if infered_qualifiers:
+                    record.qualifiers = infered_qualifiers
```
Williambraecky: I think this could be easier to read and maintain (untested)
```suggestion
            qualifiers = defaultdict(set)
            key_count = defaultdict(int)
            for content in record.error_content_ids:
                for key, value in content.qualifiers.dict.items():
                    qualifiers[key].add(value)
                    key_count[key] += 1
            record.qualifiers = {k: v[0] for k, v in qualifiers.items() if len(v) == 1 and key_count[k] == len(record.error_content_ids)}
```

Williambraecky (commented):
Excited for the feature 🤩

runbot/models/build_error.py
```diff
@@ -103,6 +103,9 @@ class BuildError(models.Model):
     tags_min_version_id = fields.Many2one('runbot.version', 'Tags Min version', help="Minimal version where the test tags will be applied.")
     tags_max_version_id = fields.Many2one('runbot.version', 'Tags Max version', help="Maximal version where the test tags will be applied.")
 
+    common_qualifiers = JsonDictField('Selection Qualifiers', compute='_compute_common_qualifiers', store=True, help="Minimal qualifiers in common needed to link error content.")
+    similar_ids = fields.One2many('runbot.build.error', compute='_compute_similar_ids', string="Similar Errors", help="Similar Errors based on common qualifiers")
```
Xavier-Do, line 107:
```suggestion
    similar_content_ids = fields.One2many('runbot.build.error', compute='_compute_similar_content_ids', string="Similar Errors content", help="Similar Errors based on common qualifiers")
```

runbot/models/build_error.py
```diff
@@ -149,6 +152,30 @@ def _compute_random(self):
...
+            if record.common_qualifiers:
+                query = SQL(
+                    r"""SELECT error_id FROM runbot_build_error_content WHERE error_id != %s AND qualifiers @> %s""",
+                    record.id,
+                    json.dumps(record.common_qualifiers.dict),
+                )
+                self.env.cr.execute(query)
+                record.similar_ids = self.env['runbot.build.error'].browse([rec[0] for rec in self.env.cr.fetchall()])
+            else:
+                record.similar_ids = False
```
```suggestion
    def _compute_similar_content_ids(self):
        for record in self:
            if record.common_qualifiers:
                query = SQL(
                    r"""SELECT id FROM runbot_build_error_content WHERE error_id != %s AND qualifiers @> %s""",
                    record.id,
                    json.dumps(record.common_qualifiers.dict),
                )
                self.env.cr.execute(query)
                record.similar_content_ids = self.env['runbot.build.error.content'].browse([rec[0] for rec in self.env.cr.fetchall()])
            else:
                record.similar_content_ids = False
```

Xavier-Do:
We may create a view list for pages and make one compute for both fields, one method for actions, ... to avoid repetitive code.

But looks good overall, lets merge it that way we can cleanup later.

PR #1047, opened 2025-02-07
[ADD] runbot_merge: create controller to merge a commit

mart-e:
This used to be done in RPC but the goal is to be able to call this from another server and not rely on a user account.


Williambraecky:
FWIW I believe mergebot still uses the 17.0 branch

mart-e:
Oh right, I was targetting runbot at first (which is on 18) and ended up on mergebot.
Will retarget.

mart-e:
Replaced by #1054

PR #1049, opened 2025-02-10
[IMP] runbot: add tracking on error qualiliers

Xavier-Do:


runbot/views/build_error_views.xml
```diff
@@ -460,6 +460,10 @@
                 </field>
               </group>
             </sheet>
+              <div class="oe_chatter">
+                  <field name="message_follower_ids" widget="mail_followers"/>
+                  <field name="message_ids" widget="mail_thread"/>
+              </div>
```
Williambraecky: It's just `<chatter/>` now iirc

PR #1068, opened 2025-02-26
[IMP] runbot: hooks'n'stuff

Williambraecky:
- Use webhook secrets with github (optional)
- Store webhooks to be processed by the leader
- automagically match prs with runbot errors if the pr_body contains the appropriate content


runbot/models/branch.py
```diff
@@ -87,6 +87,28 @@ def _compute_reference_name(self):
...
+        """
+        self.ensure_one()
+        if not self.pr_body:
+            return
+        pr_body_regex = r'^.*runbot.*error[^\d]*(\d+)[ \S]*\r?$'
+        error_ids = re.findall(
+            pr_body_regex,
+            self.pr_body,
+            re.IGNORECASE | re.MULTILINE
+        )
```
xmo-odoo: I think the regex can be simplified to this:
```suggestion
        pr_body_regex = 
        error_ids = re.findall(
            r'\brunbot.*error.*?(\d+)',
            self.pr_body,
            re.IGNORECASE | re.ASCII
        )
```
- there's no reason to limit the pattern to one match per line
- `\d+` is greedy so there's no need to end it with something wonky, especially not "space or non-whitespace"
- no idea why `\r` is even there
- by default re will consider e.g. `២` to be a valid `\d`, which probably isn't desirable here

runbot/models/branch.py
```diff
@@ -87,6 +87,28 @@ def _compute_reference_name(self):
...
+        # We need to check if they exist and search does nothing if error_ids is empty
+        errors = self.env['runbot.build.error'].sudo().search([
+            ('id', 'in', error_ids), ('fixing_pr_id', '=', False),
+        ])
+        errors.fixing_pr_id = self
+        for error in errors:
+            error.message_post(body='Fixing pr automatically set through PR message')
```
xmo-odoo, line 149: Might be useful to log the user triggering this, and also maybe to first check that the author of the PR is a known / internal user to avoid abuse (as anyone can create a PR)

PR #1071, opened 2025-02-28
[IMP] runbot: add kanban view and stages to build error page

lse-odoo:
Introduce a kanban view to the runbot build error records.
Kanban cards show the most valuable field with relevant icons.

The concept of "state" was also introduced for the build error
model.
Currently there is 3 states:
 1. New: default state for any new error build
 2. Solved: when the issue should be solved
 3. Disabled: build error currently disabled (test-tags)

States are expected to be used by the user. For instance after fixing an underteministic error the user can change the
state from new -> solved so that the user knows that he's actively not working on it

Visuals (Note: the state "Done" have been removed):
image
image

### DIFF
--- a/runbot/__manifest__.py
+++ b/runbot/__manifest__.py
@@ -6,7 +6,7 @@
     'author': "Odoo SA",
     'website': "http://runbot.odoo.com",
     'category': 'Website',
-    'version': '5.10',
+    'version': '5.11',
     'application': True,
     'depends': ['base', 'base_automation', 'website'],
     'data': [
--- a/runbot/migrations/18.0.5.11/post-migration.py
+++ b/runbot/migrations/18.0.5.11/post-migration.py
@@ -1,0 +1,15 @@
+
+def migrate(cr, version):
+    # Build errors with test_tags are considered disabled
+    cr.execute("""
+               UPDATE runbot_build_error
+               SET state = 'disabled'
+               WHERE test_tags IS NOT NULL AND active IS TRUE
+    """)
+    # Archived build errors are considered solved
+    # Note: archived records with test-tags are considered solved too
+    cr.execute("""
+               UPDATE runbot_build_error
+               SET state = 'solved'
+               WHERE active IS FALSE
+    """)
--- a/runbot/models/build_error.py
+++ b/runbot/models/build_error.py
@@ -105,6 +105,15 @@ class BuildError(models.Model):
     error_count = fields.Integer("Error count", store=True, compute='_compute_count')
     previous_error_id = fields.Many2one('runbot.build.error', string="Already seen error")
 
+    state = fields.Selection([
+        ('new', 'New/Unsolved'),
+        ('solved', 'Solved'),
+        ('disabled', 'Disabled'),
+    ], default='new', tracking=True, group_expand=True,
+    help="New: Error is new and not yet solved.\n"
+         "Solved: Error should be solved.\n"
+         "Disabled: Error is disabled (generally set with test tags).")
+    )
     responsible = fields.Many2one('res.users', 'Assigned fixer', tracking=True)
     customer = fields.Many2one('res.users', 'Customer', tracking=True)
     team_id = fields.Many2one('runbot.team', 'Assigned team', tracking=True)
--- a/runbot/views/build_error_views.xml
+++ b/runbot/views/build_error_views.xml
@@ -5,6 +5,9 @@
       <field name="model">runbot.build.error</field>
       <field name="arch" type="xml">
         <form>
+          <header>
+            <field name="state" widget="statusbar" options="{'clickable': '1'}"/>
+          </header>
           <sheet>
               <div name="button_box">
                 <button class="oe_stat_button" type="object" icon="fa-exclamation-circle" name="action_get_build_link_record">
@@ -329,6 +332,55 @@
         <field name="binding_view_types">list</field>
     </record>
 
+    <record id="build_error_view_kanban" model="ir.ui.view">
+      <field name="name">runbot.build.error.kanban</field>
+      <field name="model">runbot.build.error</field>
+      <field name="arch" type="xml">
+          <kanban default_group_by="state" quick_create="false" default_order="last_seen_date desc">
+              <templates>
+                <t t-name="card">
+                  <widget name="web_ribbon" title="Test-tags" bg_color="bg-danger" invisible="not test_tags"/>
+                  <field name="name" class="fw-bold fs-5"/>
+                  <group>
+                    <div class="d-flex align-items-center">
+                      <i class="fa fa-clock-o me-2" title="Date interval from first seen to last seen"/>
+                      <field name="first_seen_date" widget="remaining_days"/>
+                      <i class="fa fa-long-arrow-right mx-2 oe_read_only" title="to"/>
+                      <field name="last_seen_date" widget="remaining_days"/>
+                    </div>
+
+                    <div class="d-flex align-items-center gap-1">
+                      <i class="fa fa-repeat" title="Number of occurence"/>
+                      <field name="error_count"/>
+                    </div>
+
+                    <div class="d-flex align-items-center gap-1">
+                      <i class="fa fa-bullseye" title="Triggers"/>
+                      <field name="trigger_ids" widget="many2many_tags"/>
+                    </div>
+                  </group>
+
+                  <footer>
+                    <div class="d-flex align-items-center gap-1">
+                      <field name="activity_ids" widget="kanban_activity"/>
+                    </div>
+                    <div class="d-flex align-items-center gap-1 ms-auto">
+                      <i class="fa fa-random text-danger" title="inconsistant" invisible="not random"/>
+                      <i class="fa fa-users" title="Responsible team"/>
+                      <field name="team_id"/> <i t-if="!record.team_id.raw_value">no team</i>
+                      <i class="fa fa-address-card" title="Investigator"/>
+                      <field name="customer" widget="many2one_avatar_user"/>
+                      <i class="fa fa-wrench" title="Solver"/>
+                      <field name="responsible" widget="many2one_avatar_user"/>
+                      <field name="fixing_pr_url" widget="url" text="PR" invisible="not fixing_pr_url"/>
+                    </div>
+                  </footer>
+                </t>
+              </templates>
+          </kanban>
+      </field>
+    </record>
+
     <record id="build_error_view_tree" model="ir.ui.view">
       <field name="name">runbot.build.error.list</field>
       <field name="model">runbot.build.error</field>
@@ -478,7 +530,7 @@
       <field name="name">Errors</field>
       <field name="res_model">runbot.build.error</field>
       <field name="path">error</field>
-      <field name="view_mode">list,form</field>
+      <field name="view_mode">list,kanban,form</field>
       <field name="context">{'search_default_not_fixed_errors': True, 'active_test': False}</field>
     </record>
 


lse-odoo:
Current visuals:
image

runbot/data/build_error_stage.xml
```diff
@@ -0,0 +1,18 @@
+<odoo>
+    <record model="runbot.build.error.stage" id="build_error_stage_new">
```
angv-odoo: maybe it would be also smart to have "ignored" or "test-tags" column, which would include tasks with test-tags?

runbot/models/build_error.py
```diff
@@ -76,6 +76,18 @@ def _search(self, operator, value):
...
+class BuildErrorStage(models.Model):
+    _name = 'runbot.build.error.stage'
+    _description = 'Build Error Stage'
+    _order = 'sequence'
+
+    name = fields.Char(string='Stage Name', required=True, translate=True)
+    description = fields.Text(string='Stage description', translate=True)
+    sequence = fields.Integer('Sequence', default=1)
+    fold = fields.Boolean(string='Folded in Kanban', default=False)
+
```
angv-odoo: If it is to work, we will also need to change values of "active" and "tes_tags" every time stage changes to either solved/unsolved/ignored. And vice versa, if user decides to change te value of "active" or "test_tags" fields, we stage should also change based on that. 
Otherwise, we will have to manually change the stage of every already solved ticket from unsolved to done.

lse-odoo:
few updates:
image
- added the "ignored" state before the done one
- ribbon if test-tags
- icon if it is a "random" test (no dice nor clover available for icons :( )
- added a PR link

runbot/data/build_error_stage.xml
lse-odoo: doneee

runbot/models/build_error.py
lse-odoo: I guess for some like "solved" field we should in theory remove them as they are redundant with the stages. But as a lot of other code depends on it I guess I will need to make it the same as the existing field and have a script to run at install to adapt records.
But you are right I should be more aware of this, I will rethink about it when the stages will be totally decided. For instance should issues that did not happen since X days be moved to "ignored" automatically ?

lse-odoo:
while looking if we can remove the "Open (not fixed)" field, I realized that it is technically the `active` field. So when you archive the record this field get toggled. It also explains why it is no longer visible/researchable by default.
In this case, the "Done" stage does not make a lot of sense as it will appear empty every time if we keep current structure. If not the side effect might be very big (unstable).
EDIT: as it is basically the archived state, I decided to considered it as a distinct "stage" after done. An issue after 30 days in done would automatically be archived

lse-odoo:
@angv-odoo would you mind giving it another look please ? :)

lse-odoo:
TODO: stage ignored for test-tags (constraint or put it in "ignored" stage & in migration)
EDIT: https://github.com/odoo/runbot/commit/9c4e43fba6450b53a9bf0cc269b6232b5b2abe19 for example of migration script

lse-odoo:
@angv-odoo ready fora full review if you have some time :)

runbot/security/ir.model.access.csv
```diff
@@ -151,4 +151,5 @@ access_runbot_build_stat_regex_wizard,access_runbot_build_stat_regex_wizard,mode
 
 access_runbot_host_message,access_runbot_host_message,runbot.model_runbot_host_message,runbot.group_runbot_admin,1,0,0,0
 
-
+access_runbot_build_error_stage_user,access_runbot_build_error_stage_user,runbot.model_runbot_build_error_stage,base.group_user,1,0,0,0
+access_runbot_build_error_stage_admin,access_runbot_build_error_stage_admin,runbot.model_runbot_build_error_stage,base.group_user,1,1,1,1
```
Williambraecky:
```suggestion
access_runbot_build_error_stage_admin,access_runbot_build_error_stage_admin,runbot.model_runbot_build_error_stage,runbot.group_runbot_admin,1,1,1,1
```

runbot/models/build_error.py
```diff
@@ -90,6 +90,18 @@ def _search(self, operator, value):
...
+class BuildErrorStage(models.Model):
+    _name = 'runbot.build.error.stage'
+    _description = 'Build Error Stage'
+    _order = 'sequence'
+
+    name = fields.Char(string='Stage Name', required=True)
+    description = fields.Text(string='Stage description')
+    sequence = fields.Integer('Sequence', default=1)
+    fold = fields.Boolean(string='Folded in Kanban', default=False)
```
Williambraecky: I don't think any of those fields are actually relevant and `stage_id` could be a selection field IMO

runbot/models/build_error.py
```diff
@@ -347,6 +360,7 @@ def _compute_graph(self):
 
     @api.constrains('test_tags')
     def _check_test_tags(self):
+        self.stage_id = self.env.ref('runbot.build_error_stage_ignored').id
```
Williambraecky: I don't know if it's safe to change a value within a constrains 🤔

runbot/views/build_error_views.xml
```diff
@@ -329,6 +332,59 @@
...
+                      <i class="fa fa-long-arrow-right mx-2 oe_read_only" title="to"/>
+                      <field name="last_seen_date" widget="remaining_days"/>
+                    </div>
+
+                    <div class="d-flex align-items-center gap-1">
+                      <i class="fa fa-repeat" title="Number of occurence"/>
+                      <field name="error_count"/>
+                    </div>
+
+                    <div class="d-flex align-items-center gap-1" style="display: flex; align-items: center;">
```
Williambraecky: Why are you giving it both the style and the bootstrap classes? 🤔

Williambraecky (changes requested):
Personally I am not convinced by the idea of a kanban view for this model, but to each their own, I wouldn't make it the default, that's for sure.
Thanks for your work though :)

runbot/models/build_error.py
```diff
@@ -586,6 +600,37 @@ def action_link_errors(self):
...
+        now = fields.Datetime.now()
+        build_error_stage_new = self.env.ref('runbot.build_error_stage_new').id
+        build_error_stage_solved = self.env.ref('runbot.build_error_stage_solved').id
+        build_error_stage_done = self.env.ref('runbot.build_error_stage_done').id
+
+        # Very old done error eventually get archived
+        self.search([
+            ('stage_id', '=', build_error_stage_done),
+            ('last_seen_date', '<', now - relativedelta(days=nbr_day_done_to_archive)),
+        ]).write({'active': False})
```
Xavier-Do: We don't want to have a fully automated way to archive error for now. We already thought about doing something like tracking all forwardport pr and detecting that all fixing pr are merged and that the error did not appear since it was merged, but this is really arbitrary to mark as solved if it diddn't appear for a while
- error that breaks every year, and is not fixed yet. 
- error that changed of error message because of another refactoring (need to be cleaned, merged)

Keeping it alive will allow to easily identify the cause and potential existing conversations. With this logic it would be archived.

 If you don't want to have old error as noise in your view, you can just add a filter to remove error not seen for a while. There is actually one for the opposite (disaplay olfd errors only) that we use sometime to do this operation manually, but checking first it it looks legit. (merged fixing pr, very rare, not appearing anymore and no similar error found)
 
 We are working to reduce the number of error using qualifier and adding to to help identify if the error really disappeared, this will help to do something similar in the future but much more precise. 
 
Also, we NEVER want to disable an error with a test-tag without being aware of that, it must be a manual operation followed if particular attention on the following stagings. An error may disappear just because the error message changed slightly.

runbot/models/build_error.py
lse-odoo: it can indeed be and I started it as a selection field.

But it does have its shares of drawback:
- No way to fold the stages by default (at least, to my knowledge) -> in standard it looks like there is the need for the `fold` record to have this feature
- No way to dynamically add new stages (but this can indeed be argued as it will likely not be the case and can be added in the code)
- There was some issue to show all stage; some would dissapear if there was no record with this stage (but it might be possible standard-ly with `group_expand=True`, but I'm pretty sure it didn't worked when I tried out before)

Let me know your opinion if this drawbacks are not relevant in your opinion and I will switch it as a selection field

runbot/views/build_error_views.xml
lse-odoo: I struggled so much on that part to make it looks right, I forgot to cleaned it my bad :) I'll remove the style in the next commit as it does not look to affect

runbot/models/build_error.py
lse-odoo: I will remove the archive part. The "Done" stage is enough in my opinion so that we know that the issue is likely solved. As the done stage is folded by default in the kanban view it will not bother the user. We will let you the decision of archiving them like it is currently

runbot/models/build_error.py
Williambraecky: You can do that with a custom group_expand

lse-odoo: I did the changes for a selection field. The folding part can only be done through overidding `_read_group_fill_results` if I'm not mistaken (at least on the python side

lse-odoo:
@Williambraecky could you re-review when you have some time please ? :)

runbot/views/build_error_views.xml
```diff
@@ -478,7 +534,7 @@
       <field name="name">Errors</field>
       <field name="res_model">runbot.build.error</field>
       <field name="path">error</field>
-      <field name="view_mode">list,form</field>
+      <field name="view_mode">kanban,list,form</field>
```
Williambraecky:
```suggestion
      <field name="view_mode">list,kanban,form</field>
```

runbot/__manifest__.py
```diff
@@ -18,6 +18,7 @@
         'data/runbot_data.xml',
         'data/runbot_error_regex_data.xml',
         'data/website_data.xml',
+        'data/ir_cron_data.xml',
```
Williambraecky: Remove this

runbot/data/ir_cron_data.xml
```diff
@@ -0,0 +1,10 @@
+<odoo>
```
Williambraecky: Remove the file

runbot/models/build_error.py
```diff
@@ -105,6 +105,17 @@ class BuildError(models.Model):
...
+    state = fields.Selection([
+        ('new', 'New/Unsolved'),
+        ('solved', 'Solved'),
+        ('ignored', 'Ignored'),
+        ('done', 'Done'),
+    ], default='new', tracking=True, group_expand=True,
+    help="New: Error is new and not yet solved. Build error not happening for a long time in this state will automatically move to Done\n"
+         "Solved: Error should be solved. Build error not happening for a small time in this state will automatically move to Done\n"
+         "Ignored: Error is ignored. No change of state will apply\n"
+         "Done: Error is done. No change of state will apply except if issue re-appear, then it will go back to New"
```
Williambraecky: Adapt the help message, no automation as of now.

runbot/models/build_error.py
```diff
@@ -105,6 +105,17 @@ class BuildError(models.Model):
     error_count = fields.Integer("Error count", store=True, compute='_compute_count')
     previous_error_id = fields.Many2one('runbot.build.error', string="Already seen error")
 
+    state = fields.Selection([
+        ('new', 'New/Unsolved'),
+        ('solved', 'Solved'),
+        ('ignored', 'Ignored'),
+        ('done', 'Done'),
```
Williambraecky: Since we aren't doing automation I don't think the done stage is really necessary, once an error is solved it is archived to the done column would only contain archived errors anyways.

runbot/models/build_error.py
```diff
@@ -586,6 +597,37 @@ def action_link_errors(self):
...
+        # Solved error that did not appear after a short time are marked as done
+        self.search([
+            '|', 
+            '&',
+                ('state', '=', 'new'),
+                ('last_seen_date', '<', now - relativedelta(days=nbr_day_new_to_done)),
+            '&',
+                ('state', '=', 'solved'),
+                ('last_seen_date', '<', now - relativedelta(days=nbr_day_solved_to_done)),
+        ]).write({'state': 'done'})
```
Williambraecky: Remove this

runbot/views/build_error_views.xml
```diff
@@ -329,6 +332,59 @@
...
+      <field name="name">runbot.build.error.kanban</field>
+      <field name="model">runbot.build.error</field>
+      <field name="arch" type="xml">
+          <kanban default_group_by="state" quick_create="false" default_order="last_seen_date desc">
+              <templates>
+                <t t-name="card">
+                  <widget name="web_ribbon" title="Test-tags" bg_color="bg-danger" invisible="not test_tags"/>
+                  <field name="name" class="fw-bold fs-5"/>
+                  <group>
+                    <div style="display: flex; align-items: center;">
```
Williambraecky:
```suggestion
                    <div class="d-flex align-items-center">
```

runbot/migrations/18.0.5.10/post-migration.py
```diff
@@ -0,0 +1,11 @@
+
+from odoo import api, SUPERUSER_ID
+
+
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+
+    # Archived build errors are set are considered Done
+    env['runbot.build.error'].search([['active', "=", False]]).write({'state': 'done'})
```
Williambraecky: If we remove done
```suggestion
    env['runbot.build.error'].search([['active', "=", False]]).write({'state': 'solved'})
```

runbot/migrations/18.0.5.10/post-migration.py
```diff
@@ -0,0 +1,11 @@
...
+    # Build errors with test_tags are set to Ignored
+    env['runbot.build.error'].search([['test_tags', '!=', False]]).write({'state': 'ignored'})
```
Williambraecky: I know those are very simple operations but they should honestly be done through raw sql queries.

Williambraecky (changes requested):
Hello 👋 

We are fine with the kanban view, however:
 - We do not want any automation without really thinking about it and agreeing together about what should do what, if you wish to discuss it further you are welcome to come to our office or start a thread/task about it :)
 - We do not want to make the kanban view the default view, we are very used to working with the list view and having to handle state is not something we want as of now, at least not more than what we _currently_ have.

The comments below have been made with that in mind, if you still with to have the kanban view without the automation those changes should be made/considered and we can merge the view by itself :)

Cheers,

lse-odoo:
@Williambraecky thanks for the detailed review :) I think I got carried away by the original PR idea. Let's start gentle and we will see with time if it needs automation or not you are right.
Apart from the PR/commit message that I need to adapt, do you see something coding wise to change ? :)

runbot/migrations/18.0.5.11/post-migration.py
```diff
@@ -0,0 +1,14 @@
...
+    cr.execute("""
+               UPDATE runbot_build_error
+               SET state = 'solved'
+               WHERE active = false
+    """)
+    # Build errors with test_tags are considered ignored
+    cr.execute("""
+               UPDATE runbot_build_error
+               SET state = 'ignored'
+               WHERE test_tags is not null
```
Williambraecky: To be more precise:
```suggestion
               WHERE test_tags is not null AND active IS TRUE
```

Williambraecky (approved):
Looks good to me 👍 
Thanks :)

NOTE: I haven't pulled the latest version to test but code seems quite similar so I didn't deem it necessary.

runbot/views/build_error_views.xml
```diff
@@ -329,6 +332,59 @@
...
+
+                    <div class="d-flex align-items-center gap-1">
+                      <i class="fa fa-repeat" title="Number of occurence"/>
+                      <field name="error_count"/>
+                    </div>
+
+                    <div class="d-flex align-items-center gap-1">
+                      <i class="fa fa-code-fork" title="Concerned Odoo versions"/>
+                      <field name="version_ids" widget="many2many_tags"/>
+                    </div>
```
Williambraecky: This field is causing a lot of latency as it is not stored and is a related field.
For now let's remove it :)
```suggestion
```

Williambraecky (changes requested):
Tested locally and found some issues :)

runbot/migrations/18.0.5.11/post-migration.py
```diff
@@ -0,0 +1,15 @@
+
+def migrate(cr, version):
+    # Build errors with test_tags are considered ignored
+    cr.execute("""
+               UPDATE runbot_build_error
+               SET state = 'ignored'
```
d-fence: `ignored` does not seem to be an appropriate name for errors that should be considered as a priority to fix.
Why not `disabled` or `neutralized`

runbot/models/build_error.py
```diff
@@ -586,6 +595,15 @@ def action_link_errors(self):
         base_error = self_sorted[0]
         base_error._merge(self_sorted - base_error)
 
+    @api.model
+    def _read_group_fill_results(self, domain, groupby, annoted_aggregates, read_group_result, read_group_order=None):
+        # Override to fold ignored state
```
d-fence: Why should we fold errors that have to be fixed in priority ?

d-fence (changes requested):

runbot/migrations/18.0.5.11/post-migration.py
lse-odoo: no one told me there was a priority to fix, it does indeed make sense in this case to show them and adapt the state name. Thanks for the info :)

runbot/views/build_error_views.xml
lse-odoo: should I do the same for `trigger_ids` as it's also computed not-store ?

Williambraecky: `trigger_ids` is fine, as it it computed stored, I tested locally with the diff above and perf was good 👌

PR #1073, opened 2025-03-03
[ADD] runbot: smart button on build errors for build links list

lse-odoo:
Before this commit:
The only way to have details on the builds was to use the embed one2many widget in the build error form view.
This was pretty inconvenient for complicated search as there is no way to filter the records or group them

After this commit:
A list view now exist for the model `runbot.build.error.link` allowing to filter and group on convenient fields. This list view can be accessed on the runbot build error through a smart button


lse-odoo:
Visuals:
image
image
image

PR #1078, opened 2025-03-05
[IMP] runbot: automatic error management

Xavier-Do:
Add an action to post the message for exceptions automatically as well as rebuilding the templates.

### DIFF
--- a/runbot/views/upgrade.xml
+++ b/runbot/views/upgrade.xml
@@ -47,7 +47,7 @@
                 <field name="create_uid"/>
                 <field name="team_id"/>
                 <field name="message"  readonly="0"/>
-                <button name="action_post_message" string="Post message on all pull requests" type="object" invisible="not create_date"/>
+                <button name="action_post_message" groups="runbot.group_runbot_admin" string="Post message on all pull requests" type="object" invisible="not create_date"/>
                 <button name="action_auto_rebuild" string="rebuild templates" type="object" invisible="not create_date"/>
               </group>
             </sheet>
--- a/runbot/views/upgrade.xml
+++ b/runbot/views/upgrade.xml
@@ -26,7 +26,6 @@
         </field>
     </record>
 
-
     <record id="upgrade_exception_form" model="ir.ui.view">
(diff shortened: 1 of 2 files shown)


runbot/views/upgrade.xml
```diff
@@ -36,12 +35,20 @@
...
+                </field>
                 <field name="elements"/>
                 <field name="info"/>
                 <field name="create_date"/>
                 <field name="create_uid"/>
                 <field name="team_id"/>
-                <field name="message"/>
+                <field name="message"  readonly="0"/>
+                <button name="action_post_message" string="Post message on all pull requests" type="object" invisible="not create_date"/>
+                <button name="action_auto_rebuild" string="rebuild templates" type="object" invisible="not create_date"/>
```
Williambraecky: I think those are better as smart button or as header buttons
Screenshot 2025-03-07 at 14 28 05

runbot/views/upgrade.xml
```diff
@@ -36,12 +35,20 @@
...
+                    <field name="alive"/>
+                  </list>
+                </field>
                 <field name="elements"/>
                 <field name="info"/>
                 <field name="create_date"/>
                 <field name="create_uid"/>
                 <field name="team_id"/>
-                <field name="message"/>
+                <field name="message"  readonly="0"/>
```
Williambraecky, line 49: Might as well make the field itself `readonly=False`

runbot/views/upgrade.xml
Williambraecky:
```suggestion
                <button name="action_post_message" groups="runbot.group_runbot_admin" string="Post message on all pull requests" type="object" invisible="not create_date"/>
                <button name="action_auto_rebuild" string="rebuild templates" type="object" invisible="not create_date"/>
```

Williambraecky (commented):
Noice

Xavier-Do:
Note that the current actions are more there as a test, the expected future behaviour may be to do that automatically when creating the errors once fully validated by using it manually.

Williambraecky (approved):

PR #1089, opened 2025-03-10
[IMP] runbot: add warning on some config_views

Xavier-Do:
The warning will appear on build (and children)
This is useful to explain the purpose of the build and special attention to give when reading the result. (templates)


Williambraecky (approved):
The color and the text being center makes it feel like something is wrong with the build.

That being said, I don't have an alternative for you, so lgtm.

Xavier-Do:
The goal is to get user attention but I could display that in info if the build is in success (or not display it at all)

PR #1090, opened 2025-03-10
[REF] runbot: prepare terrain for owl capabilities

Williambraecky:
- Reorganize static folder
- Fix diff_match_patch export which was borked
- Vendor owl.js and expose through odoo module
- Add a way to render owl components on the frontend
- Add a way to add interaction through colibri framework
- Rewrite runbot.js using colibri
- Remove direct rels to assets in layout template


Williambraecky:
Diff is big but this is necessary if we want modern javascript for future improvements/refactoring.

JQuery is only required by the stats page, which https://github.com/odoo/runbot/pull/1035 already covers.

PR #1093, opened 2025-03-12
[18.0][FIX] runbot: Wrong syntax in the batch template.

anhbtit:
Screenshot from 2025-03-12 11-06-26


Ngquang (approved):

Williambraecky (approved):

Williambraecky:
Thanks for the fix :)
Please note depending on your python version the previous syntax may work 👍 , it's not recommended of course.

Ngquang:
> Thanks for the fix :) Please note depending on your python version the previous syntax may work 👍 , it's not recommended of course.

Thanks mate, we tested this with 3.10 and got this error. Should note it on Document br.

PR #1097, opened 2025-03-12
[FIX] runbot: remove keepquery from navbar main

Williambraecky:
When doing a search on runbot, the main navbar item (for the current project) keeps the search, making it less convenient to arrive back at the homepage.


Williambraecky:
@Xavier-Do we might also want to remove the search from queryparams for other projects, though I see the value in keeping it as it is now.

PR #1101, opened 2025-03-13
[IMP] runbot: public API

Williambraecky:
Until now api routes on runbot have been created through custom website
pages in production.
We want to unify the API by making a 'public' api, inspired by the way
`web_search_read` works.

This commit adds:
 - A route to list all publicly available models
 - A route to do a read on a public model
 - A route to fetch the publicly available specification for a model
 - A public model mixin that provides all the tools required to support
   the above mentionned routes.

The mixin adds the ability to add the `public` attribute on fields.
Any field marked as public can then be publicly queried through the
controller.
Relational fields work in a nested manner (`fields` key in the field's
sub-specification) (up to a depth of 10).
The public api does not allow going through a relationship back and
front (parent->child->parent is NOT allowed).

Because we are based on `web_search_read`, we heavily focus on
validating the specification, for security reasons, and offset the load
of reading to the `web_read` function (we currently don't provide limit
metadata).

NOTE: this PR focuses on the public api feature, not the implementation so information might be lacking as of writing this.


runbot/models/public_model_mixin.py
```diff
@@ -0,0 +1,315 @@
...
+    'date', 'datetime', 'selection', 'jsonb',
+    'many2one', 'one2many', 'many2many',
+}
+RELATIONAL_FIELD_TYPES = {'many2one', 'one2many', 'many2many'}
+SPEC_MAX_DEPTH = 10
+SPEC_METADATA_FIELD = {
+    '__type', '__help',
+}
+DEFAULT_LIMIT = 20
+DEFAULT_MAX_LIMIT = 60
```
Williambraecky, line 27: Those two are only moved once and have been replaced with methods so probably not necessary.

runbot/models/public_model_mixin.py
```diff
@@ -0,0 +1,315 @@
...
+    fields: Dict[str, 'SubSpecification']
+Specification = Dict[str, Union[Dict, 'SubSpecification']]
+
+SUPPORTED_FIELD_TYPES = { # Perhaps this should be a list of class instead
+    'boolean', 'integer', 'float', 'char', 'text', 'html',
+    'date', 'datetime', 'selection', 'jsonb',
+    'many2one', 'one2many', 'many2many',
+}
+RELATIONAL_FIELD_TYPES = {'many2one', 'one2many', 'many2many'}
+SPEC_MAX_DEPTH = 10
```
Williambraecky, line 22: Perhaps this could also be moved to a method instead.

PR #1106, opened 2025-03-14
[IMP] build error automerge mecanism

Xavier-Do:


runbot/common.py
```diff
@@ -22,6 +22,21 @@
 dest_reg = re.compile(r'^\d{5,}-.+$')
 
 
+def transactioncache(method):
+    def wrapper(self, *args, **kwargs):
```
Williambraecky, line 28: Always good to use [functools.wraps](https://docs.python.org/3/library/functools.html#functools.wraps)
```suggestion
    @functools.wraps(method)
    def wrapper(self, *args, **kwargs):
```

runbot/common.py
```diff
@@ -22,6 +22,21 @@
...
+        cache = self.env.cr.cache
+        method_key = (self._name, method.__name__)
+        params_key = (args, frozenset(kwargs.items()))
+        # should check id key is serializable
+        if method_key not in cache:
+            cache[method_key] = {}
+        if params_key not in cache[method_key]:
+            cache[method_key][params_key] = method(self.with_context({}), *args, **kwargs)
+        return cache[method_key][params_key]
+    wrapper.clear_transaction_cache = lambda self: self.env.cr.cache.pop((self._name, method.__name__), None)
```
Williambraecky: Could you not use the method itself as a cache key?
```suggestion
        method_key = (method, )
        params_key = (args, frozenset(kwargs.items()))
        # should check id key is serializable
        if method_key not in cache:
            cache[method_key] = {}
        if params_key not in cache[method_key]:
            cache[method_key][params_key] = method(self.with_context({}), *args, **kwargs)
        return cache[method_key][params_key]
    wrapper.clear_transaction_cache = lambda self: self.env.cr.cache.pop((method, ))
```

runbot/models/build_error.py
```diff
@@ -234,8 +238,8 @@ def _compute_similar_ids(self):
         for record in self:
             if record.common_qualifiers:
                 query = SQL(
-                    r"""SELECT id FROM runbot_build_error WHERE id != %s AND common_qualifiers @> %s""",
-                    record.id,
+                    r"""SELECT id FROM runbot_build_error WHERE id not in %s AND common_qualifiers @> %s""",
+                    tuple(record.ids),
```
Williambraecky: record is insure_one here so this has no effect; we might want to move the query outside of the loop.

runbot/models/build_error.py
```diff
@@ -248,8 +252,8 @@ def _compute_similar_content_ids(self):
         for record in self:
             if record.common_qualifiers:
                 query = SQL(
-                    r"""SELECT id FROM runbot_build_error_content WHERE error_id != %s AND qualifiers @> %s""",
-                    record.id,
+                    r"""SELECT id FROM runbot_build_error_content WHERE error_id not in %s AND qualifiers @> %s""",
+                    tuple(record.ids),
```
Williambraecky: Same

runbot/models/build_error.py
```diff
@@ -262,8 +266,8 @@ def _compute_analogous_ids(self):
         for record in self:
             if record.common_qualifiers:
                 query = SQL(
-                    r"""SELECT id FROM runbot_build_error WHERE id != %s AND unique_qualifiers @> %s""",
-                    record.id,
+                    r"""SELECT id FROM runbot_build_error WHERE id not in %s AND unique_qualifiers @> %s""",
+                    tuple(record.ids),
```
Williambraecky: Same

runbot/models/build_error.py
```diff
@@ -276,8 +280,8 @@ def _compute_analogous_content_ids(self):
         for record in self:
             if record.common_qualifiers:
                 query = SQL(
-                    r"""SELECT id FROM runbot_build_error_content WHERE error_id != %s AND qualifiers @> %s""",
-                    record.id,
+                    r"""SELECT id FROM runbot_build_error_content WHERE error_id not in %s AND qualifiers @> %s""",
+                    tuple(record.ids),
```
Williambraecky: Same

runbot/models/build_error.py
```diff
@@ -563,13 +568,19 @@ def _parse_logs(self, ir_logs):
                     error.metadata = logs[0].metadata
 
                 continue
-            new_build_error_content = self.env['runbot.build.error.content'].create({
+            vals = {
```
Williambraecky: There is no need to store the vals before calling create here

runbot/tests/test_build_error.py
```diff
@@ -667,26 +686,115 @@ def test_dashboard_tile_simple(self):
...
+            #   insert
+            #   post message create
+            #   search error history (could be computed)
+            # +2 to create the first error
+
+            for i in range(10):
+                mode = chr(97 + i)
+                error_contents |= self.env['runbot.build.error.content'].create({
+                    'content': f'Tour foobar_tour failed at step click_here in mode mode_{mode}',
+                })
```
Williambraecky, line 758: Why not create in batch?

runbot/models/build_error.py
Xavier-Do: It looks like I adapted all queries without taking attention to the context, good catch

runbot/tests/test_build_error.py
Xavier-Do: Because during the nightly the error are parsed per build and not created in batch and what I'm trying to check here is the non batch creation performance for this specific use case.

runbot/models/build_error.py
```diff
@@ -660,22 +673,39 @@ def _set_error_history(self):
...
+                if field in self._fields:
+                    vals[field] = v
             if not vals.get('error_id'):
-                # TODO, try to find an existing one that could match, will be done in another pr
-                name = vals.get('content', '').split('\n')[0][:1000]
-                error = self.env['runbot.build.error'].create({
-                    'name': name,
-                })
-                vals['error_id'] = error.id
+                temp = self.new(vals)
```
Xavier-Do: try to remove new?

runbot/models/build_error.py
```diff
@@ -745,6 +787,10 @@ def _compute_similar_ids(self):
             else:
                 record.similar_ids = False
 
+    @api.depends('content', 'canonical_tag', 'module_name', 'file_path', 'function')
+    def _compute_qualifiers(self):
```
Xavier-Do: not sure this is needed since called from create

PR #1107, opened 2025-03-17
[IMP] runbot: docker image versioning

d-fence:
When one wants to update a Dockerfile with significant changes without impacting regular builds, he needs to create a test Dockerfile beforehand.

With this commit, when a Dockerfile is changed, the newly built image is tagged with a `.future` suffix, that way, the old image is still used until the image identifier is changed to the future one.

That way, the `.future` image can be tested in separately before being used in production.

### DIFF
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -145,6 +145,24 @@ def _docker_build(build_dir, image_tag):
         return dm.result
 
 
+def docker_tag(identifier_or_tag, new_tag):
+    return _docker_tag(identifier_or_tag, new_tag)
+
+def _docker_tag(identifier_or_tag, new_tag):
+    if not identifier_or_tag:
+        return
+    _logger.info('Tagging image %s to "%s"', identifier_or_tag, new_tag)
+    docker_client = docker.from_env()
+    repo, tag = new_tag.split(':')  # runbot DockerFile tags contains the repo part
+    try:
+        image = docker_client.images.get(identifier_or_tag)
+        image.tag(repo, tag)
+    except docker.errors.ImageNotFound:
+        _logger.warning('failed to find docker image with identifier %s', identifier_or_tag)
+    except docker.errors.APIError:
+        _logger.warning('failed to retag docker image with identifier %s', identifier_or_tag)
+
+
 def docker_push(image_tag, push_url='127.0.0.1:5001'):
     return _docker_push(image_tag, push_url)
 
--- a/runbot/container.py
+++ b/runbot/container.py
@@ -151,12 +151,13 @@ def docker_tag(identifier_or_tag, new_tag):
 def _docker_tag(identifier_or_tag, new_tag):
     if not identifier_or_tag:
         return
-    _logger.info('Tagging image %s to "%s"', identifier_or_tag, new_tag)
     docker_client = docker.from_env()
     repo, tag = new_tag.split(':')  # runbot DockerFile tags contains the repo part
     try:
         image = docker_client.images.get(identifier_or_tag)
-        image.tag(repo, tag)
+        if new_tag not in image.tags:
+            _logger.info('Tagging image %s to "%s"', identifier_or_tag, new_tag)
+            image.tag(repo, tag)
     except docker.errors.ImageNotFound:
         _logger.warning('failed to find docker image with identifier %s', identifier_or_tag)
     except docker.errors.APIError:
--- a/runbot/migrations/18.0.5.10/post-migration.py
+++ b/runbot/migrations/18.0.5.10/post-migration.py
@@ -1,0 +1,14 @@
+import logging
+
+_logger = logging.getLogger(__name__)
+
+def migrate(cr, version):
+    cr.execute("""
+        UPDATE runbot_dockerfile
+           SET image_identifier = subq.identifier, image_future_identifier = subq.identifier
+          FROM (SELECT DISTINCT ON (dockerfile_id) dockerfile_id, identifier
+                  FROM runbot_docker_build_result
+                 WHERE result='success' order by dockerfile_id, create_date desc)
+            AS subq
+         WHERE runbot_dockerfile.id = subq.dockerfile_id;
+        """)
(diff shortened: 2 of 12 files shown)


runbot/container.py
```diff
@@ -391,3 +409,12 @@ def sanitize_container_name(name):
     """Returns a container name with unallowed characters removed"""
     name = re.sub('^[^a-zA-Z0-9]+', '', name)
     return re.sub('[^a-zA-Z0-9_.-]', '', name)
+
+
+def docker_get_identifier(tag_or_id):
```
Xavier-Do: This doesn't look to be used anywhere

runbot/models/host.py
```diff
@@ -166,18 +166,22 @@ def _docker_update_images(self):
         else:
             _logger.info('Building docker images...')
             for dockerfile in self.env['runbot.dockerfile'].search([('to_build', '=', True)]):
-                dockerfile._build(self)
+                identifier = dockerfile._build(self)
```
Xavier-Do: We should manage the case where we have a failure and the return value may be Falsy

runbot/models/docker.py
```diff
@@ -124,7 +124,10 @@ class Dockerfile(models.Model):
 
     name = fields.Char('Dockerfile name', required=True, help="Name of Dockerfile")
     active = fields.Boolean('Active', default=True, tracking=True)
+    image_identifier = fields.Char('Identifier')
+    image_future_identifier = fields.Char('Future Identifier')
```
Xavier-Do: I would also add a computed boolean field saying if the two fields are out of sync and add a visual signal in docker list + form views

Also add an action to sync image_future_identifier into image_identifier maybe?

Xavier-Do (changes requested):
Looks promising!

runbot/models/host.py
d-fence: Done by keeping the same identifier when the return value is Falsy

runbot/models/docker.py
d-fence: - Visual signal done with a simple red ribbon and red decoration in view list
- Do we really need an action to copy an identifier from one field to the other ?

runbot/migrations/18.0.5.10/post-migration.py
```diff
@@ -0,0 +1,11 @@
+import logging
+
+_logger = logging.getLogger(__name__)
+
+def migrate(cr, version):
+    cr.execute("""
+        UPDATE runbot_dockerfile
+        SET image_identifier = subq.identifier, image_future_identifier = subq.identifier
+        FROM (SELECT dockerfile_id, identifier FROM runbot_docker_build_result WHERE result='success' order by create_date desc) AS subq
```
Williambraecky: The subquery should limit by 1 per dockerfile_id, no?
We want to get the latest identifier I assume

Williambraecky:
```suggestion
        FROM (SELECT DISTINCT ON (dockerfile_id) dockerfile_id, identifier FROM runbot_docker_build_result WHERE result='success' order by dockerfile_id, create_date desc) AS subq
```

d-fence: Sure, although the query was working by chance

runbot/models/docker.py
Xavier-Do: The action may be overkill in form view but could be interesting in list view (but we can add it latter) 
Use case: update a base image, all images will rebuild. Wait for the nightly. If everything is green: Update future for all.

runbot/models/build.py
```diff
@@ -853,6 +853,8 @@ def _docker_run(self, step, cmd=None, ro_volumes=None, **kwargs):
             ro_volumes[f'/data/build/{dest}'] = source
         if 'image_tag' not in kwargs:
             kwargs.update({'image_tag': self.params_id.dockerfile_id.image_tag})
+        if self.params_id.config_data.get('docker_use_future') and not kwargs['image_tag'].endswith('.future'):
+            kwargs['image_tag'] += '.future'
```
Williambraecky, line 857: Should it not be:
```suggestion
            kwargs['image_tag'] += self.params_id.dockerfile_id.image_future_identifier
```

Williambraecky: (or `image_future_tag`)

Williambraecky: Should be tested!!

d-fence: `kwargs['image_tag'] = self.params_id.dockerfile_id.image_future_tag`, but sure, it would be far better than hard-coding `.future`

Xavier-Do: No it is not correct, we want to be able to pass an arbitrary docker file as kwargs ;)

runbot/models/build.py
```diff
@@ -853,6 +853,8 @@ def _docker_run(self, step, cmd=None, ro_volumes=None, **kwargs):
             ro_volumes[f'/data/build/{dest}'] = source
         if 'image_tag' not in kwargs:
             kwargs.update({'image_tag': self.params_id.dockerfile_id.image_tag})
+        if self.params_id.config_data.get('docker_use_future') and not kwargs['image_tag'].endswith('.future'):
+            kwargs['image_tag'] = self.params_id.dockerfile_id.image_future_tag
```
Xavier-Do: This is not correct, we need to keep the image_tag potentially passed as kwargs. With this version step like [ruff](https://runbot.odoo.com/odoo/runbot.build/76794901/runbot.build.config/209/runbot.build.config.step/439) will fail because you would replace the specified image_tag with the future of another image.

runbot/models/build.py
Williambraecky: We should do it in the previous if, then:
if image_tag not provided: use_future ? image_tag_future : image_tag

Xavier-Do: No the whole purpose it to be able to have the future of all docker images using this flag. [Odoo:Ruff](https://runbot.odoo.com/odoo/runbot.build/76794901/runbot.build.config/209/runbot.build.config.step/439) is passed as an image_tag, with this flag we expect to use the Odoo:Ruff.future

runbot/models/docker.py
```diff
@@ -124,7 +124,10 @@ class Dockerfile(models.Model):
 
     name = fields.Char('Dockerfile name', required=True, help="Name of Dockerfile")
     active = fields.Boolean('Active', default=True, tracking=True)
+    image_identifier = fields.Char('Identifier')
```
Xavier-Do: Add tracking (both fields) (bisou)

Xavier-Do (changes requested):

PR #1108, opened 2025-03-17
[FIX] runbot: restore "error already known" message

Williambraecky:
Displays it for the error manager group as well as if you are assigned to it.


Williambraecky:
Validated by LSE

PR #1112, opened 2025-03-25
[IMP] runbot: assign teams to error from qualifiers

Williambraecky:
Uses the 'module' key from qualifier to define the team to assign if it is available, as we consider qualifiers to be more precise than file path it also takes precedence over the latter.

### DIFF
--- a/runbot/models/build_error.py
+++ b/runbot/models/build_error.py
@@ -12,7 +12,7 @@ from markupsafe import Markup
 from werkzeug.urls import url_join
 from odoo import models, fields, api
 from odoo.exceptions import ValidationError, UserError
-from odoo.tools import SQL
+from odoo.tools import SQL, lazy
 
 from ..fields import JsonDictField
 
@@ -107,7 +107,9 @@ class BuildError(models.Model):
 
     responsible = fields.Many2one('res.users', 'Assigned fixer', tracking=True)
     customer = fields.Many2one('res.users', 'Customer', tracking=True)
-    team_id = fields.Many2one('runbot.team', 'Assigned team', tracking=True)
+    team_id = fields.Many2one('runbot.team', 'Assigned team', compute='_compute_team_id', inverse='_inverse_team_id', store=True, tracking=True)
+    manual_team_id = fields.Many2one('runbot.team', 'Manually assigned team')
+    auto_team_id = fields.Many2one('runbot.team', 'Automatically assigned team', readonly=True) # This is a computed field but not really
     fixing_commit = fields.Char('Fixing commit', tracking=True)
     fixing_pr_id = fields.Many2one('runbot.branch', 'Fixing PR', tracking=True, domain=[('is_pr', '=', True)])
     fixing_pr_alive = fields.Boolean('Fixing PR alive', related='fixing_pr_id.alive')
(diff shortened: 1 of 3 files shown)


d-fence (approved):
Looks very very good to me.
🥇

runbot/models/build_error.py
```diff
@@ -497,19 +497,28 @@ def action_view_analogous_qualified_contents(self):
...
-                    if error_content.file_path:
-                        if teams is None:
-                            teams = self.env['runbot.team'].search(['|', ('path_glob', '!=', False), ('module_ownership_ids', '!=', False)])
-                            repos = self.env['runbot.repo'].search([])
-                        team = teams._get_team(error_content.file_path, repos)
-                        if team:
-                            record.team_id = team
-                            break
+        teams = lazy(self.env['runbot.team'].search, ['|', ('path_glob', '!=', False), ('module_ownership_ids', '!=', False)])
+        repos = lazy(self.env['runbot.repo'].search, [])
```
Xavier-Do, line 511: I've started experimenting in my branch with a pattern that could be used in this case too. 

in team.py and repo.py
```python
    def _get_ownership_teams(self):
        if '_get_ownership_teams' not in self.env.cr.cache:
             self.env.cr.cache['_get_error_auto_merge'] = self.env['runbot.team'].search(['|', ('path_glob', '!=', False), ('module_ownership_ids', '!=', False)])
        return  self.env.cr.cache['_get_error_auto_merge']
        
    def get_all_repo(self):
        if 'get_all_repo' not in self.env.cr.cache:
             self.env.cr.cache['get_all_repo'] = self.env['runbot.repo'].search([])
        return  self.env.get_all_repo.cache['_get_error_auto_merge']
```

It is a transaction cache but more flexible than the record one. Even if not officially supported using a decorator or so, this is the purpose of the cr.cache (kind of)

Then 

```suggestion
        teams = lazy(self.env['runbot.team']._get_ownership_teams)
        repos = lazy(self.env['runbot.repo'].get_all)
```

This will allow to make the search once for all error when scanning multiple of them during the nightly.

Xavier-Do (commented):
Looks good and nice integration in get_team :+1: 

I think that it could be the opportunity to go even further for not to much more work:

Right now we may assign a team on a random error (l10n) even if it is not correct, once the team is assigned we have no good way to know if the assignation was manual, or automatic.

The automatic assignation could be updated. Lets say the same error in l10n starts to appear in mail, maybe we could automatically reassign it. It would be easier if the automatic assignation was in a dedicated field, and the current one was the manual assignation (both could be managed with a computed field with inverse to hide the complexity) 

If manual field is set, this is the one to use. If not, use the automatic until manual edition. (could be discussed)

runbot/models/build_error.py
Xavier-Do: Note: this could be done in another pr ;)

Williambraecky:
> Looks good and nice integration in get_team 👍
> 
> I think that it could be the opportunity to go even further for not to much more work:
> 
> Right now we may assign a team on a random error (l10n) even if it is not correct, once the team is assigned we have no good way to know if the assignation was manual, or automatic.
> 
> The automatic assignation could be updated. Lets say the same error in l10n starts to appear in mail, maybe we could automatically reassign it. It would be easier if the automatic assignation was in a dedicated field, and the current one was the manual assignation (both could be managed with a computed field with inverse to hide the complexity)
> 
> If manual field is set, this is the one to use. If not, use the automatic until manual edition. (could be discussed)

It still depends on qualifying contents at their creation but should be good 👌

PR #1116, opened 2025-03-28
[IMP] runbot: log docker file identifier

Xavier-Do:
The current log of the docker file used in a branch is only the docker
tag, it is difficult to know what was the exact version used in this
build.

This commit changes that to log a link to the current identifier in use
as well as adding some pages to consult docker result from frontend

### DIFF
--- a/runbot/controllers/frontend.py
+++ b/runbot/controllers/frontend.py
@@ -692,10 +692,12 @@ class Runbot(Controller):
         '/runbot/dockerfile',
         '/runbot/dockerfile/<int:dockerfile_id>',
         '/runbot/dockerfile/tag/<string:docker_tag>',
+        '/runbot/dockerfile/result/<string:dockerfile_result_id>',
         '/runbot/dockerfile/version/<string:version>',
     ], type='http', auth='public', sitemap=False)
-    def dockerfile_content(self, dockerfile_id=None, version=None, docker_tag=None, **kwargs):
+    def dockerfile_content(self, dockerfile_id=None, version=None, docker_tag=None, dockerfile_result_id=None, **kwargs):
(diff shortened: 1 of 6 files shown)


runbot/controllers/frontend.py
```diff
@@ -704,16 +706,48 @@ def dockerfile_content(self, dockerfile_id=None, version=None, docker_tag=None,
             dockerfile = dockerfile_sudo.search([('image_tag', '=', docker_tag)])
         elif version:
             dockerfile = dockerfile_sudo.search([('version_ids.name', '=', version)])
+        elif dockerfile_result_id:
+            dockerfile_result = request.env['runbot.docker_build_result'].browse(int(dockerfile_result_id)).exists()
```
Williambraecky: Don't you need sudo here?

Williambraecky: Or will it be hidden route for admin only?

runbot/controllers/frontend.py
```diff
@@ -704,16 +706,48 @@ def dockerfile_content(self, dockerfile_id=None, version=None, docker_tag=None,
...
+            dockerfile_result = dockerfile_result_sudo.browse(dockerfile_result_id).exists()
+        else:
+            raise NotFound
+
+        if not dockerfile_result:
+            raise NotFound
+
+        dockerfile = dockerfile_result.dockerfile_id
+
+        if not dockerfile_result.dockerfile_id.public_visibility:
```
Williambraecky: Could be good to allow admins to view any dockerfile result:
```suggestion
        if not (dockerfile.public_visibility or request.env.user.has_group('runbot.group_runbot_admin')):
```

Williambraecky (commented):
I didn't check the visual of the page but it's probably good enough 👌

PR #1118, opened 2025-03-31
[IMP] runbot: dockerfile variants

Williambraecky:


runbot/models/docker.py
```diff
@@ -202,10 +220,13 @@ def _compute_dockerfile(self):
     def onchange_dockerfile(self):
         self.in_error = False
 
-    @api.depends('name')
+    @api.depends('name', 'parent_id.image_tag')
     def _compute_image_tag(self):
         for rec in self:
-            if rec.name:
+            if rec.parent_id and rec.variant_name:
+                variant_sub_tag = re.sub(r'[ /:\(\)\[\]]', '', rec.variant_name.lower())
```
Xavier-Do: Would be simpler to have a constraints on name instead of cleaning it (historical mistake)

runbot/migrations/18.0.5.11/pre-migration.py
```diff
@@ -0,0 +1,10 @@
+def migrate(cr, version):
+    # Replace dot with _
+    # Replace (space)\:\(\)\[\] with nothing
+    cr.execute(r"""
+    UPDATE runbot_dockerfile
+       SET name=REGEXP_REPLACE(
+            REPLACE(name, '.', '_'),
+            '[ /:\(\)\[\]]', '', 'g'
```
Xavier-Do, line 8: Maybe we could just use the tag ? (still replacing dots by _)

Williambraecky: Tbh the query is not complete as we disallow a lot more characters than before, this query takes care of our data mostly.
(only concerns `13.0` and `DockerSaas17.2`)
I don't think it's worth doing much more.

runbot/models/docker.py
```diff
@@ -177,20 +199,18 @@ def _compute_use_count(self):
...
-                    content = ''
-            else:
-                content = rec.layer_ids.render_layers()
+            layers = rec.layer_ids
+            if rec.parent_id:
+                layers = self.env['runbot.docker_layer'].new({
+                    'name': 'TEMP LAYER',
+                    'layer_type': 'reference_file',
+                    'reference_dockerfile_id': rec.parent_id.id
+                }) + layers
```
Xavier-Do: Maybe that this could be enough, to test

```suggestion
            if rec.parent_id:
                layers = rec.parent_id.layers + layers
```

PR #1152, opened 2025-06-13
Hide success sub-builds

xmo-odoo:
Useful in large split builds where 90% of sub-builds are successful and hunting down the unsuccessful ones is a pain in the ass e.g. single app test.


runbot/templates/build.xml
```diff
@@ -53,6 +53,7 @@
                    t-attf-class="{{'' if prev_ko else 'disabled '}}btn btn-default fa fa-angle-double-left"></a>
                 <a t-att-href="prev_bu.build_url" role="button" t-attf-title="Previous {{prev_bu.display_name}}"
                    t-attf-class="{{'' if prev_bu else 'disabled '}}btn btn-default fa fa-chevron-left"></a>
+                <a role="button" class="btn btn-default fa fa-compress" onclick="document.documentElement.classList.toggle('hide-success');"/>
```
d-fence, line 56: Maybe add a title that says what the button does
```suggestion
                <a role="button" class="btn btn-default fa fa-compress" onclick="document.documentElement.classList.toggle('hide-success');" title="Hide successful sub-builds"/>
```

d-fence (commented):
Nice idea. A hint on the button would be nice.

PR #1161, opened 2025-06-19
[IMP] runbot: add a button to create missing trigger slots

d-fence:


runbot/models/batch.py
```diff
@@ -499,6 +499,36 @@ def _log(self, message, *args, level='INFO'):
...
+                'dockerfile_id': dockerfile_id.id,
+                'commit_link_ids': [(6, 0, self.commit_link_ids.ids)],
+            }
+            params = self.env['runbot.build.params'].create(params_value)
+            self.env['runbot.batch.slot'].create({
+                'batch_id': self.id,
+                'trigger_id': trigger.id,
+                'params_id': params.id,
+                'link_type': 'created',
+            })
```
Xavier-Do, line 531: I wonder if this logic shouldn't be refactored with the prapare logic

PR #1171, opened 2025-07-15
[IMP] OverrideRights: add helper method to extract partner_ids from commands

vib-adhoc:
Before this PR, the create and write methods assumed that partner_ids would always be passed using a Command.SET. This caused issues when other types of Odoo commands (e.g., Command.LINK, Command.CREATE, etc.) were used — leading to errors during override rights assignment (see attached gif).

### Refactoring for partner ID extraction:

* [`runbot_merge/models/res_partner.py`](diffhunk://#diff-17930401fd68f04bea38c6138bce8862b47bf1e9c93985d01589f2f5a9d118e1R175-R192): Added a new helper method `get_partner_ids` to encapsulate the logic for extracting partner IDs from Odoo commands. This method handles various command types (`SET`, `LINK`, `UNLINK`) and ensures cleaner code.

### Updates to `create` and `write` methods:

* [`runbot_merge/models/res_partner.py`](diffhunk://#diff-17930401fd68f04bea38c6138bce8862b47bf1e9c93985d01589f2f5a9d118e1R198-R199): Updated the `create` method to use the new `get_partner_ids` helper for extracting partner IDs, replacing the inline logic with a call to the helper method.
* [`runbot_merge/models/res_partner.py`](diffhunk://#diff-17930401fd68f04bea38c6138bce8862b47bf1e9c93985d01589f2f5a9d118e1R209-L195): Modified the `write` method to use the `get_partner_ids` helper for handling partner ID updates, ensuring consistency and reducing code duplication.

Steps to reproduce the error

### DIFF
--- a/runbot_merge/models/res_partner.py
+++ b/runbot_merge/models/res_partner.py
@@ -4,7 +4,6 @@ from email.utils import parseaddr
 
 from markupsafe import Markup
 
-import odoo.tools
 from odoo import fields, models, tools, api, Command
 
 from .. import github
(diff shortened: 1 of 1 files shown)


vib-adhoc:
@xmo-odoo Hello, I have seen you active in this repository, could you take a look at this PR?

xmo-odoo:
> This caused issues when other types of Odoo commands (e.g., Command.LINK, Command.CREATE, etc.) were used — leading to errors during override rights assignment (see attached gif).

In what context does that happen? As far as I can tell from the gif and testing this pops up when *creating* an override from a partner's form, but that seems to be more error prone than useful: there likely shouldn't be tons of overrides being created all the time, so it probably would make more sense to *only* create override rights via the overrides list, and update the form to forbid creation of overrides via this field.

Alternatively (or alongside that) by the same logic that override rights shouldn't be created very commonly trying to reify the partners list by hand is probably an unnecessary optimisation:

- on create we can just get the list of partners from the records we just created
- on write we can read a map of `override:partners` before write, read the same after write, and diff the two

vib-adhoc:
Hello @xmo-odoo my apologies for the delay in getting back to you. I really appreciate you updating this to get the write working correctly.

However, I've found that there's still an issue with the create method. The problem also happens when creating overrides from the list view. I've added a new commit that fixes this, following the suggestion from your previous message.

Let me know what you think!

xmo-odoo:
Yup, I should have thought to look at it, thanks.

PR #1172, opened 2025-07-18
[IMP] runbot: faketime offset for children

d-fence:
In order to make the "faketime" builds more realistic, a time offset is added to the children builds. This offset is based on the difference between the parent build creation date and the child build creation date.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -11,6 +11,7 @@ import time
 import uuid
 
 from collections import defaultdict
+from dateutil import parser
 from pathlib import Path
 from psycopg2 import sql
 from psycopg2.extensions import TransactionRollbackError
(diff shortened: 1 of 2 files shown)


runbot/models/build.py
```diff
@@ -1125,6 +1126,9 @@ def _cmd(self, python_params=None, py_version=None, local_only=True, sub_command
 
         faketime = []
         if faketime_params := self.params_id.config_data.get('faketime'):
+            if self.parent_id:
+                parent_time_offset = self.create_date - self.parent_id.create_date
```
Xavier-Do: In a case where we have a lot of queue (nightly) the test could take a long time to start (bot main build and child)

This could lead to varying times for self.create_date - self.parent_id.create_date

Maybe that build_start  would be more reliable.
```suggestion
                parent_time_offset = self.create_date - self.parent_id.build_start  
```

But this is till not working for rebuild since the create date will be the one of the rebuild

```suggestion
                parent_time_offset = self.parent_id.build_end  - self.parent_id.build_start  
```

(actually, the build_time but expressed as a time delta.)

Unfortunately this won't work if the parent build is not finished (installing base) 

```suggestion
                parent_time_offset = (self.parent_id.build_end or self.create_date) - self.parent_id.build_start  
```

This should be close enough I guess.

PR #1195, opened 2025-08-12
[IMP] runbot: speedup history graph rendering

Xavier-Do:
Since #1189 the rendering of the page is slower, mainly because of the number of element in the svg. This commit replaces the svg with a canvas element, which is much faster to render.

This also allow to customize the rendering.

While on it, adding a possibility to get all first batch, and all last batch


Xavier-Do:
Some tweak to do regarding the cache but lets merge it as it is for now.

PR #1202, opened 2025-08-28
18.0 faketime propagate offset xdo

Xavier-Do:


runbot/models/build.py
d-fence, line 1098: Maybe get rid of this last `now` and thus remove it from the import too.
While at it bring order into the import mess.

d-fence (changes requested):

PR #1217, opened 2025-09-11
[IMP] runbot: more detailed generic failure

Xavier-Do:
Add the end of the log file for generic failure messages

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -16,7 +16,7 @@ from pathlib import Path
 from psycopg2 import sql
 from psycopg2.extensions import TransactionRollbackError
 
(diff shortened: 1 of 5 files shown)


runbot/models/build.py
```diff
@@ -678,6 +678,14 @@ def _get_docker_name(self):
         self.ensure_one()
         return '%s_%s' % (self.dest, self.active_step.name)
 
+    def _get_error_tail_message(self, log_path):
+        lines = tail(log_path)
+        details = ''
+        if lines:
+            last_lines = '\n'.join(lines)
```
d-fence: readlines returns lines with a `\n`  at the end already.

```suggestion
            return '\n' + ''.join(lines)
```

d-fence (changes requested):

PR #1223, opened 2025-09-19
[IMP] runbot_merge: recombine remaining batches after identifying a culprit PR

MiquelRForgeFlow:
**Background (current behavior):**
When a staging with multiple batches fails, runbot_merge splits the set in two halves and stages them in sequence. Only the failing half is split again (bisected); a passing half is staged and merged as a unit. Eventually, if bisecting reaches a single-batch staging that fails, the culprit PR is identified and marked as failed. After that point, the remaining PRs from the original split group continue through the existing split queue (halves and further splits only if those fail). There is no step to recombine all unaffected PRs into a single staging once the culprit is known.

**Problem:**
Even when there is only one bad PR, the remaining good PRs may still be processed across several stagings (depending on how the queued splits unfold), leading to extra CI cycles and slower throughput.

**Proposed change:**
As soon as a culprit PR is identified in a single-batch staging:

1. Mark that PR as failed (existing behavior).
~~2. Cancel any pending stagings from the same split source and remove sibling splits for that source.~~
3. Create a new split containing all remaining active batches from the original group (excluding the culprit and closed/merged ones).
4. Let the scheduler stage this recombined batch together.

**Why this helps:**
- Fewer CI runs: In the common “one culprit” scenario, we avoid cascading stagings of separate halves.
- Faster merges: Good PRs get merged together promptly.
- Safe: If another culprit or a problematic interaction exists, the recombined staging will fail and the standard bisect logic will isolate it.

**Implementation notes:**

- New helper `recombine_remaining_after_culprit(pr)` on `runbot_merge.stagings`. Finds sibling splits by `source_id`, filters remaining active batches, cancels pending sibling stagings, unlinks sibling splits for the same source, then creates a single recombined split with the remaining batches.
- Set `_order = 'id desc'` on `runbot_merge.split` so the newly created recombined split is picked first by `try_staging()`.

**How to test (manual):**
1. Prepare 4 batches `B1..B4` on the same target. Make `B1` fail in isolation; others pass.
2. Trigger a staging with all 4 → it fails → it is split into `[B1,B2]` and `[B3,B4]`.
3. Process the left split; bisect isolates `B1` as culprit (single-batch failure).
4. Verify logs show: “recombine remaining N batches …” and that a new split is created with `B2,B3,B4`.
5. Next scheduler run should stage `B2,B3,B4` together; CI passes → they merge together.
6. Repeat with a “two culprits” scenario to confirm the process repeats until all bad PRs are isolated.

**Backward compatibility & failure modes:**
- Worst case (multiple culprits/interactions): behavior naturally falls back to standard bisect.
- No change to approvals/labels; only scheduling is optimized.
- Recombined staging creation is idempotent per split source cycle.

If you’d like, I can add a short “Changelog entry”.

### DIFF
--- a/runbot_merge/models/pull_requests.py
+++ b/runbot_merge/models/pull_requests.py
@@ -2304,6 +2304,37 @@ class Stagings(models.Model):
         })
         return True
 
+    def recombine_remaining_after_culprit(self, culprit_pr):
+        """Regroup remaining active batches on the target into a single split,
+        excluding the culprit batch (single-batch path).
+        """
+        self.ensure_one()
+        culprit_batch = self.batch_ids[:1]
+
+        Split = self.env['runbot_merge.split']
+        extant_splits = Split.search([
+            ('target', '=', self.target.id),
+        ])
+        if not extant_splits:
+            _logger.info("No extant splits on target %s; skipping recombine", self.target.name)
+            return False
+
+        candidate_batches = extant_splits.mapped('batch_ids').filtered(lambda b: b.id != culprit_batch.id and b.active)
+        if len(candidate_batches) <= 1:
+            _logger.info("Nothing to recombine (remaining=%s) on %s after culprit %s",
+                         len(candidate_batches), self.target.name, culprit_pr.display_name)
+            return False
+        extant_splits.unlink()
+        new_split = Split.create({
+            'target': self.target.id,
+            'source_id': (self.parent_id or self).id,
+            'batch_ids': [Command.link(b.id) for b in candidate_batches],
+        })
+        _logger.info(
+            "Recombined %d batches after culprit %s into split %s (target=%s)",
+            len(candidate_batches), culprit_pr.display_name, new_split, self.target.name
+        )
+
     def try_splitting(self):
         batches = len(self.batch_ids)
         if batches > 1:
@@ -2357,11 +2388,12 @@ class Stagings(models.Model):
                 viewmore = ' (view more at %(target_url)s)' % status
             if pr:
                 self.fail("%s%s" % (reason, viewmore), pr)
+                self.recombine_remaining_after_culprit(pr)
             else:
                 self.fail('%s on %s%s' % (reason, head.commit_id.sha, viewmore))
             return False
 
-        # the staging failed but we don't have a specific culprit, fail
+        # the staging failed, but we don't have a specific culprit, fail
         # everything
         self.fail("unknown reason")
 
@@ -2527,6 +2559,7 @@ class Stagings(models.Model):
 
 class Split(models.Model):
     _name = _description = 'runbot_merge.split'
+    _order = 'id desc'
 
     target = fields.Many2one('runbot_merge.branch', required=True)
     source_id = fields.Many2one('runbot_merge.stagings', required=True)


MiquelRForgeFlow:
@Xavier-Do @d-fence Hope you like it.

Xavier-Do:
Hello @MiquelRForgeFlow 

Thanks for contributing

This is more a topic for @xmo-odoo but it looks like it wouldn't work well with random errors.

MiquelRForgeFlow:
@Xavier-Do to be more safe, I removed the cancelling of any pending staging from the same split source and sibling splits for that source. The new combined split should be executed regardless because the `_order` set on `runbot_merge.split`.

BTW, I asked ChatGPT how to handle better the random errors (flakes) and it proposes two ways:
1. Culprit confirmation with a short retry (“canary retry”): Before marking a PR as failed (when you detect it as single-batch), make a unique retry of that single-batch.
2. Unique retry of the recombined / single-batched when there is no clear culprit.

runbot_merge/models/pull_requests.py
```diff
@@ -2304,6 +2304,41 @@ def fail(self, message, prs=None):
...
 
+    def recombine_remaining_after_culprit(self, culprit_pr):
+        """Combines all the remaining batches of the same split group
+        (same source_id) except for the batch of the culprit PR and the closed ones,
+        creating a single Split with the rest in order to test them together.
+        """
+        culprit_batch = culprit_pr.batch_id
+        if not culprit_batch:
+            _logger.info("No culprit batch found for PR %s; skipping recombine", culprit_pr)
+            return False
```
xmo-odoo: This does not make sense, the staging only accesses the PRs through batches.

runbot_merge/models/pull_requests.py
```diff
@@ -2304,6 +2304,41 @@ def fail(self, message, prs=None):
...
+
+        Split = self.env['runbot_merge.split']
+        source = (self.parent_id or self)
+
+        sibling_splits = Split.search([('source_id', '=', source.id)])
+        if not sibling_splits:
+            _logger.info("No sibling splits for root staging %s; skipping recombine", source)
+            return False
+
+        candidate_batches = sibling_splits.mapped('batch_ids').filtered(lambda b: b.id != culprit_batch.id and b.active)
```
xmo-odoo: Splits are transient and deleted when staged
https://github.com/odoo/runbot/blob/cd13f03bda397cb6b3cdd6e3c3e222513c423315/runbot_merge/models/stagings_create.py#L93-L97
so a staged batch can not be part of a split. And if extant splits are regrouped into a new split, the old splits need to be deleted

runbot_merge/models/pull_requests.py
```diff
@@ -2304,6 +2304,41 @@ def fail(self, message, prs=None):
...
+            return False
+
+        new_split = Split.create({
+            'target': self.target.id,
+            'source_id': source.id,
+            'batch_ids': [Command.link(b.id) for b in candidate_batches],
+        })
+        _logger.info(
+            "Recombined %d batches after culprit PR#%s into %s (source=%s)",
+            len(candidate_batches), culprit_pr.number, new_split, source
```
xmo-odoo: A PR number is wholly insufficient to identify the PR, as they're only unique within a repository. The mergebot overrides `PullRequests.display_name` to write out the fully qualified short form of the PR (`{owner}/{name}#{number}`)

runbot_merge/models/pull_requests.py
```diff
@@ -2304,6 +2304,41 @@ def fail(self, message, prs=None):
...
+            _logger.info("No culprit batch found for PR %s; skipping recombine", culprit_pr)
+            return False
+
+        Split = self.env['runbot_merge.split']
+        source = (self.parent_id or self)
+
+        sibling_splits = Split.search([('source_id', '=', source.id)])
+        if not sibling_splits:
+            _logger.info("No sibling splits for root staging %s; skipping recombine", source)
+            return False
```
xmo-odoo: Almost certainly unnecessarily complicated, Split.parent_id only exists to create a link between stagings, which itself only exists with the idea that we might one day use it to try and better identify staging false positives (and auto-scan them or something).

At a higher level of abstraction, if there are splits for a target they all come from the same original staging.

runbot_merge/models/pull_requests.py
```diff
@@ -2304,6 +2304,41 @@ def fail(self, message, prs=None):
...
+        new_split = Split.create({
+            'target': self.target.id,
+            'source_id': source.id,
+            'batch_ids': [Command.link(b.id) for b in candidate_batches],
+        })
+        _logger.info(
+            "Recombined %d batches after culprit PR#%s into %s (source=%s)",
+            len(candidate_batches), culprit_pr.number, new_split, source
+        )
+        return True
```
xmo-odoo: Although the return value of `try_splitting` is not used anyway and should probably be removed, this is incorrect. The semantics of `try_splitting` is that it returns `True` if it managed to split the staging. Since the recombination of extant splits can only come at the consequence of a single-batch staging, the staging can not be split.

runbot_merge/models/pull_requests.py
```diff
@@ -2357,11 +2392,15 @@ def try_splitting(self):
                 viewmore = ' (view more at %(target_url)s)' % status
             if pr:
                 self.fail("%s%s" % (reason, viewmore), pr)
+                try:
+                    return self.recombine_remaining_after_culprit(pr)
+                except Exception:
```
xmo-odoo: Why?

xmo-odoo (commented):
I added both my understanding of the original version (a depth-first search of a culprit PR) and this version (keep the breadth-first search but recombine as soon as a "culprit" is found) to a small simulator I worked on during the odoo days. I found the original to be way better than I was assuming, while the simple recombination is not really an improvemend over the baseline version:

title              |  total | merged | failed | nondeterministic
-------------------|--------:|--------:|--------:|------------------:
batched            | 236000 | 233380 |   2222 |              398
batched_rejoin     | 238608 | 236005 |   2210 |              393
batched_depthfirst | 254902 | 252245 |   2377 |              280

(this is with a bunch of assumptions as the simulator is pretty simplistic, and over 5 years of simulation-time).

However,

- the branch for the mergebot is 17.0
- the mergebot has a fair number of tests, and as both this and the previous significantly impact the scheduling of splits I doubt the various tests which involve such splits still pass with this

MiquelRForgeFlow:
@xmo-odoo comments attended.

MiquelRForgeFlow:
@xmo-odoo is there something else to do?

xmo-odoo:
Finally got around to merging this, kinda, as well as the original form (doing depth-first search of failing PRs).  Thanks.

I removed much of the logging as I think this version is excessive, and put the recombination in `fail`: it doesn't matter that we can't really identify the specific PR in a multi-PR batch, for this it matters that we have failed a single batch and assume that batch is the deterministic source of errors. Same for timeout errors, if a batch deterministically causes timeouts then the rest is absolved.

I also put in a tunable so that the splits can just be deleted instead of recombined, if false positive rates are low then this should mostly yield higher throughput, and staging priority instability should mostly be fixed since 6fa190b08f63123cadf30afd84ab6651d30032fe so keeping splits around for that purpose (not having a batch overtaken by a latter one on restage) should mostly be unnecessary.

PR #1224, opened 2025-09-22
[FIX] runbot: explicitly define http_interface in all cases

Xavier-Do:


d-fence:
Needed after [this PR](https://github.com/odoo/odoo/pull/226569) as the request comes through the Docker network interface.

PR #1237, opened 2025-10-21
[IMP] runbot: add a button to disable regular triggers

d-fence:
[IMP] runbot: add a button to disable regular triggers

It's sometimes useful to disable regular triggers on a bundle to only
starts manual or custom triggers.

With this commit, a button is added on the bundle page to disable
regular triggers by creating disabled custom triggers.


runbot/views/bundle_views.xml
```diff
@@ -83,7 +84,7 @@
                                 <list editable="bottom">
                                     <field name="start_mode"/>
                                     <field name="use_base_commits"/>
-                                    <field name="trigger_id" domain="[('project_id', '=', parent.project_id)]"/>
+                                    <field name="trigger_id" domain="[('project_id', '=', parent.project_id), ('category_id.name', '=', 'Default')]"/>
```
Xavier-Do, line 87: not ideal but lets move

PR #1251, opened 2025-11-22
[19.0][FIX] runbot_merge: Wrong module import and remove extra argument on method override

anhbtit:
# What is this PR for ?
- Fix wrong module import
- Remove extra argument on method override


xmo-odoo:
Thanks for the contribution, however the mergebot is currently only compatible and tested with Odoo 17. As such I'd rather not have 19.0-specific contributions causing conflicts during an eventual migration and updated.

PR #1253, opened 2025-11-25
[REF] runbot: tracking_value revamp

pparidans:
Globally, a cleanup of the tracking value diff implementation:
- avoid multiple (many?) diff recompute using memoization
- uses CopyButton component
- modernize a bit, the JavaScript syntax
- "Boostrapify" and refine diff rendering
- diff's line highlight on hover is back

### DIFF
--- a/runbot/static/src/js/fields/tracking_value.js
+++ b/runbot/static/src/js/fields/tracking_value.js
@@ -1,74 +1,78 @@
 /** @odoo-module **/
 import { patch } from "@web/core/utils/patch";
 import { Message } from "@mail/core/common/message";
+import { CopyButton } from "@web/core/copy_button/copy_button";
+import { memoize } from "@web/core/utils/functions";
+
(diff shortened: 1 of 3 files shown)


runbot/static/src/js/fields/tracking_value.js
```diff
@@ -1,74 +1,68 @@
...
-            var diff_type = diffs[x][0];
-            var data = diffs[x][1];
-            var data_lines = data.split('\n');
-            for (var line_index in data_lines) {
-                var line = data_lines[line_index];
-                line = line.replace(/&/g, '&amp;');
-                line = line.replace(/</g, '&lt;');
-                line = line.replace(/>/g, '&gt;');
-                //text = text.replace(/\n/g, '<br>');
-                //text = text.replace(/ /g, '&nbsp&nbsp');
```
pparidans: Not sure why it was still done. Looks like it's escaped anyway (in `mail` ?) :thinking: 
To test...

PR #1265, opened 2025-12-02
[FIX] runbot: make github login case insensitive in team membership check

Xavier-Do:
Github login are not case sensitive, and the case is not always the same depending on where we get them from (github api, user input, etc).

Making the check case insensitive to avoid issues.

### DIFF
--- a/runbot/models/build_config_codeowner.py
+++ b/runbot/models/build_config_codeowner.py
@@ -134,7 +134,7 @@ class ConfigStep(models.Model):
                 pr = pr_by_commit[commit_link]
                 new_reviewers = reviewers - set((pr.reviewers or '').split(','))
                 if new_reviewers:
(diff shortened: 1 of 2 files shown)


runbot/models/build_config_codeowner.py
```diff
@@ -134,7 +134,7 @@ def _run_codeowner(self, build):
                 pr = pr_by_commit[commit_link]
                 new_reviewers = reviewers - set((pr.reviewers or '').split(','))
                 if new_reviewers:
-                    author_skippable_teams = skippable_teams.filtered(lambda team: team.skip_team_pr and team.github_team in new_reviewers and pr.pr_author in team._get_members_logins())
+                    author_skippable_teams = skippable_teams.filtered(lambda team: team.skip_team_pr and team.github_team in new_reviewers and pr.pr_author.lower() in team._get_members_logins())
```
d-fence: Could break when the pr_author is False. I know that there is always a pr_author for a long time but we certainly have old PR without an author set. Also, it might happen when there are github issues 🤔

PR #1269, opened 2025-12-03
[IMP] runbot: remove unused jQuery.browser file

pparidans:


pparidans:
Will be removed when jQuery is removed as a whole.

PR #1278, opened 2025-12-17
[REF] runbot: use Component in Interactions

pparidans:
Status: blocked.


pparidans:
Integrated into https://github.com/odoo/runbot/pull/1256

PR #1281, opened 2025-12-18
[IMP] runbot: add link to project's "next freeze"

pparidans:
Add the "next freeze" tag field on runbot's project and the link to the
current project's "next freeze" tag in the frontend's navbar.

### DIFF
--- a/runbot/templates/utils.xml
+++ b/runbot/templates/utils.xml
@@ -91,6 +91,11 @@
                                                 </a>
                                             </li>
                                         </t>
+                                        <li t-if="project and project.next_freeze_tag_id" class="nav-item" title="Freeze">
+                                            <a class="nav-link" t-attf-href="/runbot/{{slug(project)}}/bundle/tag/{{slug(project.next_freeze_tag_id)}}">
+                                                <i class="fa fa-snowflake-o"/>
+                                            </a>
+                                        </li>
                                         <t t-call="runbot.build_errors_link"/>
                                         <li class="nav-item dropdown" t-ignore="true">
                                             <a href="#" class="nav-link dropdown-toggle" data-bs-toggle="dropdown">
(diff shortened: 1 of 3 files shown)


runbot/controllers/frontend.py
```diff
@@ -908,3 +909,17 @@ def versions(self, **kwargs):
...
+        '/runbot/next-freeze'
+        ], website=True, auth='public', type='http', sitemap=False)
+    def next_freeze(self, **kwargs):
+        bundle_tags = self.env['runbot.bundle.tag'].sudo().search([])
+        icp = self.env['ir.config_parameter'].sudo()
+        is_base_regex = icp.get_param('runbot.runbot_is_base_regex', r'^((master)|(saas-)?\d+\.\d+)$')
+        bundle_tags = bundle_tags.sorted(lambda t: re.search(is_base_regex, t.name).group(1) if re.search(is_base_regex, t.name) else '00.0', reverse=True)
+        next_freeze_tag = bundle_tags[0] if bundle_tags else None
+        if not next_freeze_tag:
+            raise NotFound
```
xmo-odoo:
```suggestion
        if not bundle_tags:
            raise NotFound

        bundle_tags = bundle_tags.sorted(lambda t: m[1] if (m := re.search(is_base_regex, t.name)) else '00.0', reverse=True)
        next_freeze_tag = bundle_tags[0]
```

runbot/controllers/frontend.py
```diff
@@ -908,3 +909,19 @@ def versions(self, **kwargs):
             'keys': ['os_version', 'python_version', 'chrome_version', 'psql_version', 'pip_freeze', 'packages'],
         }
         return request.render('runbot.versions', qctx)
+
+    @route([
+        '/runbot/next-freeze'
+        ], website=True, auth='public', type='http', sitemap=False)
```
Xavier-Do: The page is not accessible for public user so this page should not be public and the icon hidden for non runbot user

runbot/templates/utils.xml
```diff
@@ -90,6 +90,11 @@
                                                 </a>
                                             </li>
                                         </t>
+                                        <li class="nav-item" title="Freeze">
```
Xavier-Do: We may want to have a way to toggle this icon (opt in on project ?): it makes no sense for most runbot instances, and outside freeze.
An alternative solution could be to define a bundle tag for freeze on project.

Xavier-Do (changes requested):
Small remarks

PR #1283, opened 2025-12-18
[IMP] runbot: add authors and teams on bundles

d-fence:
With this commit, the authors involved in a bundle are computed. Authors are found based on github logins first, then from the commit authors and commiters and finally from the ngram extracted from the bundle name.

The teams are infered from the authors found.

Finally an `Owning team` is automatically choosen (the first team of the teams) or can be manually set.

### DIFF
--- a/runbot/models/bundle.py
+++ b/runbot/models/bundle.py
@@ -1,8 +1,14 @@
 import datetime
 import re
-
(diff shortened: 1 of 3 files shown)


d-fence:
⚠️  After The Freeze ⚠️ 🧇

PR #1285, opened 2025-12-22
[FIX] runbot: allow upgrade_exception_manager to scan errors

Xavier-Do:
This was not working because by default an action needs to have write access on the record, adding the group 'upgrade_exception_manager' to the action sovles the issue by making the chck more specific allowing to run the action with write access on the build.

### DIFF
--- a/runbot/views/upgrade.xml
+++ b/runbot/views/upgrade.xml
@@ -6,6 +6,7 @@
          <field name="binding_model_id" ref="runbot.model_runbot_build" />
         <field name="type">ir.actions.server</field>
         <field name="state">code</field>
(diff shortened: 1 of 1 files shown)


runbot/views/upgrade.xml
```diff
@@ -6,6 +6,7 @@
          <field name="binding_model_id" ref="runbot.model_runbot_build" />
         <field name="type">ir.actions.server</field>
         <field name="state">code</field>
+        <field name="group_ids" ref="runbot.group_runbot_upgrade_exception_manager"/>
```
d-fence:
```suggestion
        <field name="group_ids" eval="[Command.link([ref('runbot.group_runbot_upgrade_exception_manager')])]"
```

d-fence (changes requested):

PR #1289, opened 2025-12-28
19.0 upgrade matrix followup xdo

Xavier-Do:
This pr aims to simplify the upgrade logic in the long term by using upgrade matrix instead of of defining how to behave on each step

Even if it should be possible to configure it as before (using a complement, and upgrades in separate builds) this also open the possibility to change how the upgrades are organized

Example in 19.0

Before:

```
--|Template build 19 (install database)
  |-- upgrade 19 -> 19.1
  |-- upgrade 19 -> master
  
--| upgrade build 19
  |-- upgrade 18 -> 19
  |-- upgrade 18.4 -> 19
```

After

```
--| Template build 19 (install database)

--| Test templates 19 (Trigger depends on  Template build)
  |-- upgrade 19 -> 19.1
  |-- upgrade 19 -> master
  
--| upgrade build 19
  |-- upgrade 18 -> 19
  |-- upgrade 18.4 -> 19
```

Or even better 
```
--| Template build 19 (install database)
  
--| upgrade build 19 (Trigger depends on  Template build)
  |-- upgrade 18 -> 19
  |-- upgrade 18.4 -> 19
  |-- upgrade 19 -> 19.1
  |-- upgrade 19 -> master
```

### DIFF
--- a/runbot/models/batch.py
+++ b/runbot/models/batch.py
@@ -400,8 +400,12 @@ class Batch(models.Model):
                 'create_batch_id': self.id,
                 'used_custom_trigger': bool(trigger_custom),
             }
-
-            params_value['builds_reference_ids'] = trigger._reference_builds(self)
+            if not (trigger.upgrade_step_id.upgrade_matrix_id or trigger.upgrade_dumps_trigger_id.upgrade_step_id.upgrade_matrix_id):
+                # when set the build dynamicaly gets upgrade reference builds,
+                # and needs to be batch dependant since the potentially tested build
+                # could come from the current batch
+                # TODO remove upgrade cleanup
+                params_value['builds_reference_ids'] = trigger._reference_builds(self)
 
             params = self.env['runbot.build.params'].create(params_value)
 
--- a/runbot/tests/test_upgrade.py
+++ b/runbot/tests/test_upgrade.py
@@ -29,6 +29,7 @@ class TestUpgradeFlow(RunbotCase):
             'name': 'upgrade',
             'project_id': self.project.id,
             'manifest_files': False,
+            'upgrade_paths': 'migrations',
         })
         self.remote_upgrade = self.env['runbot.remote'].create({
             'name': 'bla@example.com:base/upgrade',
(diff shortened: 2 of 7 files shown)


runbot/tests/test_upgrade.py
```diff
@@ -326,7 +327,7 @@ def assertOk(build, from_build, target_build, db_suffix):
         assertOk(b_173_master_demo, self.template_per_version['saas-17.3'], master_upgrade_build, 'all')
         assertOk(b_173_master_no_demo, self.template_per_version['saas-17.3'], master_upgrade_build, 'no-demo-all')
 
-        self.assertEqual(b_17_master_demo.params_id.commit_ids.repo_id, self.repo_odoo | self.repo_upgrade | self.repo_enterprise)
+        self.assertEqual(b_17_master_demo.params_id.commit_ids.repo_id, self.repo_upgrade)
```
Xavier-Do, line 330: Before this pr, all commits here passed but what will actually be used are the commit of the source and target builds. 

In some case, the source or target can be the parent one, containing all commits but the upgrade step only actually use upgrade commits from the current build. This changes removes useless commits to allow for a better param fingerprint and improved clarity

runbot/models/batch.py
```diff
@@ -400,8 +400,12 @@ def _fill_missing(branch_commits, match_type):
...
                 'used_custom_trigger': bool(trigger_custom),
             }
-
-            params_value['builds_reference_ids'] = trigger._reference_builds(self)
+            if not (trigger.upgrade_step_id.upgrade_matrix_id or trigger.upgrade_dumps_trigger_id.upgrade_step_id.upgrade_matrix_id):
+                # when set the build dynamicaly gets upgrade reference builds,
+                # and needs to be batch dependant since the potentially tested build
+                # could come from the current batch
+                # TODO remove upgrade cleanup
+                params_value['builds_reference_ids'] = trigger._reference_builds(self)
```
Xavier-Do, line 408: The upgrade builds were rarely linked because the _reference_builds will vary quite often (new commit in one of the versions) meaning that even if some of the subuilds are identical to previous ones, everything will rebuild since the build changed at the slot level.

After this pr upgrade builds are indirectly batch defendant, allowing to 
- remove the references from the params and using the one stored on the batch
- allowing to reference build that don't exist yet in the future.

The link will be done at the sub-build level (not ideal but ... )

PR #1292, opened 2026-01-05
[FIX] runbot: click/middle-click on FrontendUrl fields

pparidans:
Before this commit, and since the update to Odoo 19.0, click/middle-click on FrontendUrl fields in list view opened the record's modal form view in addition to the actual link in a new tab.

This commit fixes it by stopping the event's propagation beyond the actual `<a>` tag.

Steps to reproduce:
- open an error's form view,
- in the Error Content list view, click on the first/last seen date
- related build link opens in a new tab => the error's form view also opens in modal


pparidans:
Similar fix on its way for standard url field: https://github.com/odoo/odoo/pull/247815

PR #1310, opened 2026-01-30
[IMP] runbot: add a wizard to set pr on errors

d-fence:
A simple runbot user is not allowed to set PR on runbot build errors to avoid mess in build errors. So, most of the time when a user wants to set a fixing PR on a build error, he needs to send us a message.

With this commit, an error responsible can set a PR by using a wizard on his own errors.


d-fence:
not needed -- see #1316 instead

PR #1349, opened 2026-03-13
[FIX] runbot_merge: ensure compatibility with Python 3.10

vib-adhoc:
The use of backslashes inside f-string expressions was introduced in Python 3.12. In older versions (like Python 3.10), this triggers a SyntaxError: "f-string expression part cannot include a backslash".

First commit fixes the issue by moving the string with the newline character to a separate variable before using it in the f-string, ensuring compatibility with older Python environments.

The 'Self' type hint was introduced in the 'typing' module in Python 3.11. To maintain compatibility with Python 3.10, this commit adds a fallback to 'typing_extensions'.

Second commit prevents an ImportError when loading the 'ir.cron' and 'runbot_merge.batch.validate' model in
environments running Python 3.10.

### DIFF
--- a/runbot_merge/models/crons/validate.py
+++ b/runbot_merge/models/crons/validate.py
@@ -1,7 +1,10 @@
 import contextlib
 import json
 import logging
-from typing import Self
+try:
+    from typing import Self
+except ImportError:
+    from typing_extensions import Self
 
 from odoo import fields, models, api
 
--- a/runbot_merge/models/ir_cron/__init__.py
+++ b/runbot_merge/models/ir_cron/__init__.py
@@ -2,7 +2,11 @@ import contextvars
 import datetime
 import math
 import time
-from typing import Self, Literal
+from typing import Literal
+try:
+    from typing import Self
+except ImportError:
+    from typing_extensions import Self
 
 from odoo import models
 
--- a/runbot_merge/models/pull_requests.py
+++ b/runbot_merge/models/pull_requests.py
@@ -1949,10 +1949,11 @@ For your own safety I've ignored *everything in your entire comment*.
         (h, out, err, hh) = conflicts.get(previous_pr) or (None, None, None, None)
         if h:
             sout = serr = ''
+            newline_ellipsis = '\n[...]'
             if out.strip():
-                sout = f"\nstdout:\n```\n{utils.shorten(out, 8096, '\n[...]')}\n```\n"
+                sout = f"\nstdout:\n```\n{utils.shorten(out, 8096, newline_ellipsis)}\n```\n"
             if err.strip():
-                serr = f"\nstderr:\n```\n{utils.shorten(err, 8069, '\n[...]')}\n```\n"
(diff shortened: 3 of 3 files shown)


vib-adhoc:
Hi @xmo-odoo, could you please consider this change so we can continue using Python 3.10 in our current infrastructure?

Please note that I don't have the necessary permissions to view the 'ci/runbot' and 'ci/style' failure details, but I am fully available to make any further adjustments or fixes if needed.

Best regards

xmo-odoo:
> could you please consider this change so we can continue using Python 3.10 in our current infrastructure?

I don't really mind, but it's definitely going to break again in the future as mergebot.odoo.com runs on 3.12 so that's what I use to run tests when I work on it.

FWIW I've also been assuming more and more that there is no reason to use it outside odoo's peculiar requirements, since public offerings (hosted bors, gitlab, github, graphite, mergify, ...) should work for the overwhelming majority of use cases, even if the CI is bespoke.

Also you can ignore the statuses, they're just for the runbot and completely outdated (since the runbot now works off of the  19.0 branch).

vib-adhoc:
Thanks for the feedback and for considering the PR! I completely understand that the project targets Python 3.12 and that future updates might break compatibility with 3.10 again. We'll take responsibility for maintaining our side as needed.

Regarding your suggestion about other tools, we'll definitely keep your comment in mind. However, beyond the some customizations we've built on top, we find runbot_merge to be a very powerful tool in general that fits our workflow perfectly. That's why we prefer staying with it for now and adapting to the changes you introduce.

On that note, I have a quick question: For how long do you plan to use version 17 yourselves? We've already upgraded to newer Python versions for v18 and v19, but we’ve kept our v17 infrastructure on 3.10 for now. Knowing when you plan to move away from v17 would help us decide whether to invest time in upgrading Python for this specific environment or just wait to migrate to a later version when you do.

Thanks again for your help!

xmo-odoo:
> On that note, I have a quick question: For how long do you plan to use version 17 yourselves? We've already upgraded to newer Python versions for v18 and v19, but we’ve kept our v17 infrastructure on 3.10 for now. Knowing when you plan to move away from v17 would help us decide whether to invest time in upgrading Python for this specific environment or just wait to migrate to a later version when you do.

I've no real plans on that front, I need to move up eventually as the maintenance window is closing and infra is going to breathe down my neck, but at the same time 17.0 on 3.12 has done fine by the mergebot. I should probably update the environment by odoo days / end of year.

Dropping bootstrap from the frontend is actually part of that (as well as better performances / faster feeling): the mergebot has a fair amount of customizations especially to support dark mode on a bootstrap version which didn't support it, and I don't want to have to re-do them on top of 5.3 (and hit conflicts from *its* dark mode support), which odoo was migrated to in 17.3 (the runbot's workaround was to vendor its own copy of bootstrap, but bootstrap is a huge dependency for the mergebot's needs, so removing it makes a lot more sense)

vib-adhoc:
Thanks a lot for the information and for your time! This really helps us with our planning.

Best regards

PR #1369, opened 2026-03-30
[FIX] runbot: btn-default styling, cleanup & fix active state

pparidans:
BS button "default" variant styling coherence depending on the states (regular/hover/active) and fixing the missing border color on active state.

| Before | After |
|--------|--------|
| image | image | 
| image| image|


d-fence (approved):
LGTM let's see if xdo agrees

PR #1374, opened 2026-04-02
[IMP] runbot: module dependant/dependencies

Xavier-Do:


runbot/models/build.py
```diff
@@ -1120,6 +1123,43 @@ def _get_modules_to_test(self, modules_patterns=''):
...
+        return sorted(search(modules, depth))
+
+    def _get_modules_dependencies(self, modules, depth=None):
+        self.ensure_one()
+        dependency_graph, dependant_graph = self._dependency_graph()
+        return self.search_modules_graph(modules, dependency_graph, depth)
+
+    def _get_dependant_modules(self, modules, depth=None):
+        dependency_graph, dependant_graph = self._dependency_graph()
+        return self.search_modules_graph(modules, dependant_graph, depth)
```
d-fence: As suggested by Ruff in the style build, prefix the unused variables with '_' or use '_' instead.

```suggestion
        dependency_graph, _ = self._dependency_graph()
        return self.search_modules_graph(modules, dependency_graph, depth)

    def _get_dependant_modules(self, modules, depth=None):
        _, dependant_graph = self._dependency_graph()
        return self.search_modules_graph(modules, dependant_graph, depth)
```

d-fence (changes requested):
Fix [style](https://runbot.odoo.com/runbot/batch/2455600/build/106250018). Otherwise LGTM.

PR #1382, opened 2026-04-09
[IMP] runbot: handle main command exit status

d-fence:
Since the runbot executes Docker containers non-blockingly, there is no direct way to capture the main command's exit status.

This commit writes the exit status code to a file at the end of the main Docker command. The feature is now used in Semgrep builds.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -1032,6 +1032,8 @@ class BuildResult(models.Model):
         starting_config = self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_default_odoorc')
         if isinstance(cmd, Command):
             rc_content = cmd.get_config(starting_config=starting_config)
+            if step.check_exit_status:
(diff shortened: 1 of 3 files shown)


runbot/models/build.py
```diff
@@ -1042,6 +1032,7 @@ def _docker_run(self, step, cmd=None, ro_volumes=None, env_variables=None, **kwa
         starting_config = self.env['ir.config_parameter'].sudo().get_param('runbot.runbot_default_odoorc')
         if isinstance(cmd, Command):
             rc_content = cmd.get_config(starting_config=starting_config)
+            cmd.finals = [['echo', r'\$0', '>', f'/data/build/logs/{step.sanitized_name(self)}_exit_status.txt']] + cmd.finals
```
Xavier-Do:
```suggestion
            cmd.finals = [['echo', r'\$?', '>', f'/data/build/logs/{step.sanitized_name(self)}_exit_status.txt']] + cmd.finals
```

d-fence: lol, nice catch

PR #1387, opened 2026-04-16
[REF] runbot: use message queue for kill requests

Xavier-Do:
This will be a first step to test the message queue before switching state recomputation inside it.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -43,7 +43,7 @@ from ..fields import JsonDictField
 
 _logger = logging.getLogger(__name__)
 
-result_order = ['ok', 'warn', 'ko', 'skipped', 'killed', 'manually_killed']
+result_order = ['ok', 'warn', 'ko', 'skipped', 'killed', 'manually_killed']  # TODO remove manually_killed
 state_order = ['pending', 'testing', 'waiting', 'running', 'done']
 
 COPY_WHITELIST = [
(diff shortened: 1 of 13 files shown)


runbot/models/build.py
```diff
@@ -1232,38 +1230,44 @@ def truncate(message, maxlenght=300000):
             'line': '0',
         })
 
-    def _kill(self, result=None):
+    def _kill(self, result='killed'):
```
Xavier-Do, line 1233: The default was None, only used when killing a running build, looks safer to have a default on killed.

runbot/models/build.py
```diff
@@ -1232,38 +1230,44 @@ def truncate(message, maxlenght=300000):
...
         if result:
-            v['local_result'] = result
-        build.write(v)
-
-    def _ask_kill(self, lock=True, message=None):
-        # if build remains in same bundle, it's ok like that
-        # if build can be cross bundle, need to check number of ref to build
-        if lock:
-            self.env.cr.execute("""SELECT id FROM runbot_build WHERE parent_path like %s FOR UPDATE""", ['%s%%' % self.parent_path])
+            build.local_result = result
```
Xavier-Do, line 1247: We can remove the requested_action update since it was moved in the block that will removed later

While on it, lets modernize the code (don't use write)

PR #1395, opened 2026-04-30
[FIX] runbot: use most recent similar build for quick result

Sylsee:
When looking for an existing done build to reuse via allow_similar_build_quick_result, the oldest matching build was selected. The most recent done build is more likely to carry the correct status for the current build params.

### DIFF
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -1027,7 +1027,9 @@ class ConfigStep(models.Model):
                     )
 
                     if self.allow_similar_build_quick_result:
-                        existing_done_build = next((build for build in child.params_id.build_ids.sorted('id') if build.global_state == 'done' and build.local_result not in ('skipped', 'killed')), None)
+                        existing_done_build = next((build for build in child.params_id.build_ids.sorted('id') if build.global_state == 'done' and build.global_result == 'ok'), None)
+                        if not existing_done_build:
+                            existing_done_build = next((build for build in child.params_id.build_ids.sorted('id') if build.global_state == 'done' and build.local_result not in ('skipped', 'killed')), None)
                         if existing_done_build:
                             child._log('', 'A similar [build](%s) has been found, marking as done directly', existing_done_build.build_url, log_type='markdown')
                             child.local_state = 'done'


Xavier-Do:
The strategy you propose is the same as the one used by _create_build to link an existing build on a slot

This feature is just a quick hack to reduce the load, and link build similarly to what slots will link a build on a batch, but we are limited by the result computation so only taking **done** build. 

With your suggestion, it could lead to a using build that was create after the first one (while the first one was still testing) but from a dev branch. We have no guarantee that it will be the best one.

Taking the first one ensure to have a good reproductability, all branches will point to the same build. 

This will change with https://github.com/odoo/runbot/pull/1303 where a build will be able to have multiple parent, allowing to link a testing build to replace the `allow_similar_build_quick_result` feature completly (I would like to merge that in the following weeks)

With this new strategy, it will make sense since we can link pending and testing build, meaning that we should have only one build per param except for rebuild. And rebuilding will potentially update all references (if the previous was red ? )

Also, allow_similar_build_quick_result is only used for upgrades where in theory the result is quite deterministic, it is rare that a rebuild changes the result. 

Do you have an example of where it could be useful? If yes, we can imagine to take try to sort them on the result to have success first.

Sylsee:
Thanks for the detailed explanation

I do have a concrete case where the current “first done build” picks a worse candidate: [build 109203054](https://runbot.odoo.com/runbot/batch/2502471/build/109203054) matched [build 109164638](https://runbot.odoo.com/runbot/build/109164638) (failed) even though [build 109173686](https://runbot.odoo.com/runbot/build/109173686) (success) exists.

If you’re OK with it, I can push a small adjustment that keeps the deterministic ordering but, when there are multiple `done` candidates, prefers a `done/success` one over `done/failed` (and otherwise falls back to the current “first created” choice)

runbot/models/build_config.py
```diff
@@ -1027,7 +1027,7 @@ def get_reference_builds_for_versions(versions):
                     )
 
                     if self.allow_similar_build_quick_result:
-                        existing_done_build = next((build for build in child.params_id.build_ids.sorted('id') if build.global_state == 'done' and build.local_result not in ('skipped', 'killed')), None)
+                        existing_done_build = next((build for build in child.params_id.build_ids.sorted('id', reverse=True) if build.global_state == 'done' and build.local_result not in ('skipped', 'killed')), None)
```
Xavier-Do:
```suggestion
                        existing_done_build = next((build for build in child.params_id.build_ids.sorted('id') if build.global_state == 'done' and build.global_result == 'ok'), None)
                        if not existing_done_build:
                            existing_done_build = next((build for build in child.params_id.build_ids.sorted('id') if build.global_state == 'done' and build.local_result not in ('skipped', 'killed')), None)
```

Sylsee:
@Xavier-Do up :)

Xavier-Do:
Looks good we will try that tomorrow with @d-fence

PR #1399, opened 2026-05-06
[IMP] runbot: update test tags version bounds on manual parse

d-fence:
When parsing build errors, it may happen that the error appears in a version lower or higher than the current test tags bounds.

With this commit, when a build error log is manually parsed (via the web interface or the server action), the min and max version bounds of the test tags are updated accordingly to include the new builds versions.

### DIFF
--- a/runbot/models/build_error.py
+++ b/runbot/models/build_error.py
@@ -549,6 +549,17 @@ class BuildError(models.Model):
             self.tags_min_version_id = min(self.version_ids, key=lambda rec: rec.number)
             self.tags_max_version_id = max(self.version_ids, key=lambda rec: rec.number)
 
+    def _update_version_tags(self):
+        for error in self:
+            if not (error.test_tags and error.version_ids):
+                continue
+            new_min = min(error.version_ids, key=lambda rec: rec.number)
+            new_max = max(error.version_ids, key=lambda rec: rec.number)
+            if not error.tags_min_version_id or new_min.number < error.tags_min_version_id.number:
+                error.tags_min_version_id = new_min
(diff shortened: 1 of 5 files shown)


runbot/models/build_error.py
```diff
@@ -845,6 +845,14 @@ def _parse_logs(self, ir_logs):
                         'log_date': rec.create_date,
                     })
 
+        if update_tags:
+            errors_to_update = self.env['runbot.build.error']
+            for build_error_content in build_error_contents:
+                if build_error_content.error_id and build_error_content.error_id.test_tags:
+                    errors_to_update |= build_error_content.error_id
+            for error in errors_to_update:
+                error._onchange_test_tags()
```
Xavier-Do: The onchange is not safe in this case, because it could actually make the range of covered version slower.

Lets imagine I scan in 19.0, the disable from 18.0 to master

If I have a n error in master and scan it, it will change the min from 18 to 19, renabling the test in 18.0. 

Note that the same problem could occur when modifying a test tag

It would be safer to either have a custom logic there, or take the current tags_min_version_id and tags_max_version_id in the onchange.

runbot/models/build_error.py
```diff
@@ -845,6 +856,14 @@ def _parse_logs(self, ir_logs):
                         'log_date': rec.create_date,
                     })
 
+        if update_tags:
+            errors_to_update = self.env['runbot.build.error']
+            for build_error_content in build_error_contents:
+                if build_error_content.error_id and build_error_content.error_id.test_tags:
+                    errors_to_update |= build_error_content.error_id
```
```suggestion
            errors_to_update = build_error_contents.error_id.filtered('test_tags')
```

Xavier-Do (changes requested):

PR #1401, opened 2026-05-07
[FIX] runbot: apply cpu limit multiplier in all cases

Xavier-Do:
Before this change the cpu_limit_factor was only applied in _run_install_odoo, if the cpu limit commes from the params

We also want to apply this factor in all cases, if the cpu limit comes from the step (default).

### DIFF
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -596,6 +596,10 @@ class ConfigStep(models.Model):
                 max_timeout = int(self.env['ir.config_parameter'].get_param('runbot.runbot_timeout', default=10000))
                 docker_params['cpu_limit'] = min(self.cpu_limit, max_timeout)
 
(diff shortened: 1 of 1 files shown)


runbot/models/build_config.py
```diff
@@ -596,6 +596,10 @@ def _run_step(self, build, **kwargs):
                 max_timeout = int(self.env['ir.config_parameter'].get_param('runbot.runbot_timeout', default=10000))
                 docker_params['cpu_limit'] = min(self.cpu_limit, max_timeout)
 
+            config_data = {**kwargs.get('config_data', {}), **build.params_id.config_data}
+            if config_data.get('cpu_limit_factor'):
+                docker_params['cpu_limit'] = int(docker_params['cpu_limit'] * min(float(config_data['cpu_limit_factor']), 2))
```
Xavier-Do, line 601: min to avoid mistackes, 2 looks like a reasonable limit

PR #1402, opened 2026-05-07
[IMP] runbot: replace keep_running by gc_running_date

d-fence:
The `keep_running` boolean field on BuildResult was a simple flag to prevent builds from being automatically killed. The drawback is that sometimes those builds are forgotten and stays alive forever.

This commit replaces it with a `gc_running_date` field.

Now, if a date is set, the build can only be killed by _gc_running after that date.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -360,7 +360,7 @@ class BuildResult(models.Model):
     build_url = fields.Char('Build url', compute='_compute_build_url', store=False)
     build_error_link_ids = fields.One2many('runbot.build.error.link', 'build_id')
     build_error_ids = fields.Many2many('runbot.build.error', compute='_compute_build_error_ids', string='Errors')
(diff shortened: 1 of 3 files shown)


runbot/models/build.py
```diff
@@ -360,7 +360,7 @@ class BuildResult(models.Model):
     build_url = fields.Char('Build url', compute='_compute_build_url', store=False)
     build_error_link_ids = fields.One2many('runbot.build.error.link', 'build_id')
     build_error_ids = fields.Many2many('runbot.build.error', compute='_compute_build_error_ids', string='Errors')
-    keep_running = fields.Boolean('Keep running', help='Keep running', index=True)
+    gc_running_date = fields.Date('GC Running Date', help='Running build cannot be killed before this date', index=True)
```
Xavier-Do: A partial index would be nice

Xavier-Do (changes requested):
LGTM for the rest

PR #1417, opened 2026-05-27
[IMP] runbot: add a timeout on git ssh connections

d-fence:
In order to mitigate GitHub fetch time issue, a ConnectTimeout argument is passed to ssh through git command line configuration.


d-fence:
After more investigations, it appears that the ssh connection works and hangs after that.

PR #1427, opened 2026-06-11
[IMP] runbot: update repos once a day

d-fence:
Since the runbot builders are only fetching individual commits, those fetches can last for a long time. It appears that fetching the whole repository periodically speeds up the individual fetches.

With this commit, the repo are fetched alongside with the git gc once a day.

### DIFF
--- a/runbot_builder/tools.py
+++ b/runbot_builder/tools.py
@@ -114,9 +114,14 @@ class RunbotClient():
             _logger.info('Starting git gc on repositories')
             commands = []
             host_name = self.host.name
-            for repo in self.env['runbot.repo'].search([]):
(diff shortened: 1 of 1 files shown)


runbot_builder/tools.py
```diff
@@ -130,6 +130,10 @@ def git_gc(self):
             for message in messages:
                 self.env['runbot.runbot']._warning(message)
 
+            # update the repos once a day to speed up individual commit fetches
+            for repo in self.env['runbot.repo'].search([]):
```
Xavier-Do:
```suggestion
            for repo in self.env['runbot.repo'].search([('mode', '!=', 'disabled')])
```

d-fence: while at it, let's filter them for GC too

PR #1428, opened 2026-06-11
[FIX] runbot: fix teams bundles done count

d-fence:
When counting bundles done, we need to differentiate bundles without any PR from bundles with closed PR's.

With this commit, we count the done bundles for a team using the same filter as the count for all bundles.

### DIFF
--- a/runbot/controllers/frontend.py
+++ b/runbot/controllers/frontend.py
@@ -104,7 +104,7 @@ class Runbot(Controller):
                 domain.append(('no_build', '=', False))
 
             if has_pr is not None:
-                domain.append(('has_pr', '=', bool(has_pr)))
+                domain.append(('has_active_pr', '=', bool(has_pr)))
 
             filter_mode = request.httprequest.cookies.get('filter_mode', 'default')
             if filter_mode == 'sticky':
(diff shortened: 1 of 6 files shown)


runbot/controllers/frontend.py
```diff
@@ -910,11 +910,9 @@ def bundles_by_tag(self, bundle_tag_id=None, project=None, **kwargs):
         bundles_by_team = defaultdict(list)
         nb_bundles_done = 0
```
pparidans: not needed anymore

runbot/controllers/frontend.py
```diff
@@ -910,11 +910,9 @@ def bundles_by_tag(self, bundle_tag_id=None, project=None, **kwargs):
...
         bundles = self.env['runbot.bundle'].search([('tag_ids', 'in', bundle_tag_id.id)])
+        nb_bundles_done = len(bundles.filtered(lambda b: b.has_pr and not b.has_active_pr))
```
pparidans, line 912: not sure what's best between here and in the template as a `<t t-set="..."/>` 🤔

pparidans (approved):

PR #1430, opened 2026-06-11
[IMP] runbot_merge: add callback on patch application

dylankiss:
The Weblate sync server uses this to know when to unlock its components.

Related: https://github.com/odoo/internal/pull/4231

### DIFF
--- a/runbot_merge/models/patcher.py
+++ b/runbot_merge/models/patcher.py
@@ -21,6 +21,7 @@ from email.utils import parseaddr
 from typing import Union
 
 from markupsafe import Markup
+import requests
 
 from odoo import models, fields, api
 from odoo.exceptions import ValidationError
@@ -144,6 +145,16 @@ class Patch(models.Model):
     repository = fields.Many2one('runbot_merge.repository', required=True, tracking=True)
     target = fields.Many2one('runbot_merge.branch', required=True, tracking=True)
     commit = fields.Char(size=40, string="commit to cherry-pick, must be in-network", tracking=True)
+    callback_url = fields.Char(
+        help="If set, a POST request is made to this URL after the patch was either applied or failed.",
+        tracking=True,
+    )
+    callback_retries_left = fields.Integer(
+        help="Number of times the callback can still be retried.",
+        default=5,
+        tracking=True,
+    )
+    success = fields.Boolean(help="Indicates whether the patch was successfully applied.", default=False, tracking=True)
 
     patch = fields.Text(string="unified diff to apply", tracking=True)
     format = fields.Selection([
@@ -215,7 +226,11 @@ class Patch(models.Model):
         super()._auto_init()
         self.env.cr.execute("""
         CREATE INDEX IF NOT EXISTS runbot_merge_patch_active
-            ON runbot_merge_patch (target) WHERE active
+            ON runbot_merge_patch (target) WHERE active;
+        CREATE INDEX IF NOT EXISTS runbot_merge_patch_callback
+            ON runbot_merge_patch (callback_url, active)
+         WHERE callback_url IS NOT NULL
+           AND active IS FALSE;
         """)
 
     @api.model_create_multi
@@ -354,7 +369,9 @@ class Patch(models.Model):
                 )
             else:
                 info['target_head'] = c
+                patch.success = True
 
+        self.env.ref('runbot_merge.ir_cron_fire_patch_callbacks')._trigger()
         return True
 
     def _apply_commit(self, r: git.Repo, parent: str) -> str:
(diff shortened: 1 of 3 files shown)


runbot_merge/models/patcher.py
```diff
@@ -352,11 +358,38 @@ def _apply_patches(self, target: Branch) -> bool:
...
+            'jsonrpc': '2.0',
+            'id': None,
+            'method': 'call',
+            'params': {
+                'repository': self.repository.name,
+                'branch': self.target.name,
+                'patch': self.id,
+                'success': success,
+            },
+        }
```
xmo-odoo: Unnecessarily complicated, this has no reason to be jsonrpc it should just be an HTTP POST to the specified URL (which is what the help text hints at), and since the callback is per-patch anyway it should just add the success state to the query parameters: the caller can set whatever state they need in the URL they set.

runbot_merge/models/patcher.py
```diff
@@ -335,6 +340,7 @@ def _apply_patches(self, target: Branch) -> bool:
                     # TODO: do better
                     body=e.args[0] if isinstance(e.args[0], Markup) else Markup("<pre>{}</pre>").format(e),
                 )
+                patch._fire_callback(success=False)
```
xmo-odoo: Needs to move out of line to a cron job.

runbot_merge/models/patcher.py
```diff
@@ -352,11 +358,38 @@ def _apply_patches(self, target: Branch) -> bool:
...
+            'method': 'call',
+            'params': {
+                'repository': self.repository.name,
+                'branch': self.target.name,
+                'patch': self.id,
+                'success': success,
+            },
+        }
+        try:
+            res = requests.post(self.callback_url, json=data, auth=SaasAuth(), timeout=30)
```
xmo-odoo: 30 seconds timeout is way too long.

xmo-odoo (changes requested):

runbot_merge/models/patcher.py
dylankiss: Regarding jsonrpc, I just took the same approach as the `/i18n/merge_commit` route in Mergebot. I suppose that doesn't require jsonrpc either then?

Anyway, you're right indeed that it can be made simpler by already passing the necessary state in the URL. Adapting it.

runbot_merge/models/patcher.py
dylankiss: Just to be sure: you want me to trigger a cronjob that does this instead (so it doesn't block the flow)?

runbot_merge/models/patcher.py
dylankiss: Yeah, probably. It's because the other side is also doing a request to Weblate to unlock components. I can probably move that to a cronjob as well. What timeout would you suggest?

xmo-odoo: Github webhooks require a response within 10s, that seems pretty reasonable.

runbot_merge/models/patcher.py
xmo-odoo: Indeed. A cron on a queue which should probably just take the callback URL and something like a JSON payload of additional parameters (that way if the callback queue is reused for other services that works out of the box).

runbot_merge/models/patcher.py
xmo-odoo: > Regarding jsonrpc, I just took the same approach as the `/i18n/merge_commit` route in Mergebot. I suppose that doesn't require jsonrpc either then?

Strictly speaking it doesn't but that's an internal communication between odoo services (and very specifically between the translation server and the mergebot) and that's how mat wrote it, whereas this is more of a general purpose extension.

Incidentally something else I thought of yesterday is that `SaasAuth` is not necessary either, for authentication of the hook / callback the requesting service can set some sort of random one-time code (which can be the entirety of the state on its side). The callback system is not tied to tx, or to contacting internal odoo services.

dylankiss:
@xmo-odoo I tried to rework the flow as to your suggestions.
I added an additional field to allow retries (e.g. for transient failures) and process them every 6 hours.
Related adaptation of the translations sync server: https://github.com/odoo/internal/pull/4327

dylankiss:
@xmo-odoo Small reminder. It'd great if you could have another look 🙏

xmo-odoo:
FWIW I reworked it locally to use the mergebot's retryable queue crons instead of being bespoke, I also changed the value from a direct-serialized python boolean to an integer (1 for success, currently 0 for failure but technically codes -- probably negative -- could be sent for more specific failures if that's of use).

I'll close the PR when it's deployed (unless I forget).

PR #1432, opened 2026-06-16
[IMP] runbot: create database inside docker

Xavier-Do:
To allow to run postgres in docker we cannot make any maintenance on database from outside the docker anymore, meaning that the database creations needs to be done in the docker.

### DIFF
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -428,8 +428,8 @@ class ConfigStep(models.Model):
     dockerfile_id = fields.Many2one('runbot.dockerfile', string='Dockerfile')
     dockerfile_variant = fields.Char('Docker Variant')
     # install_odoo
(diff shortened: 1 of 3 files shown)


runbot/models/build_config.py
```diff
@@ -777,8 +777,6 @@ def _run_install_odoo(self, build, config_data=None):
         db_suffix = config_data.get('db_name') or (build.params_id.dump_db.db_suffix if not self.create_db else False) or self._get_db_name(build)
         db_suffix = re.sub(r'[^a-z0-9\-_]', '_', db_suffix.lower())
         db_name = '%s-%s' % (build.dest, db_suffix)
-        if modules_to_install and self.create_db:
-            build._local_pg_createdb(db_name)
```
Xavier-Do, line 796: This is not really needed since runbot will pass the db-template in the config. 
There are also no known case where we want to automatically drop a database with the same name silently, should be done explicitly.

PR #1440, opened 2026-07-02
[IMP] runbot: improve runbot slot visuals

Xavier-Do:
The main motivation is to make the difference between testing and done build more obvious, especially when in failure

In the first batch it may look like the Enterprise Test (yellow) is finished, even if it is not the case. The only visual clue is that the cog is spinning. 

This is made more obvious with two changes: 
- The cog menu will appear blue if the build is pending/testing/waiting. 
- the match type icon is replace with a state/result icon. The icon represent the state (spinner) when the build is still running and replaced by an icon giving the result once done (making it more obvious for color blind people, in addition to the custom theme)
- the match type information is transferred to the trigger name, italic and blue when it is a link to an existing build.

Before 

image

After

image

### DIFF
--- a/runbot/static/src/css/runbot.css
+++ b/runbot/static/src/css/runbot.css
@@ -4,6 +4,7 @@
 	--btn-default-bg: var(--bs-body-bg);
 	--btn-default-border: #ccc;
 	--active-project-color: #777;
+	--text-color-link: #507279;
 }
 
 /*
@@ -29,6 +30,7 @@
 :root[data-bs-theme=dark] {
 	--btn-default-border: #333;
 	--active-project-color: #CCC;
+	--text-color-link: #97d4e0;
 }
 
 [data-bs-theme=legacy] .text-bg-info {
@@ -277,6 +279,11 @@ body, .table {
 .slot_button_group {
 	display: flex;
 	padding: 0 1px;
+}
+
+.slot_link_matched > .slot_name {
+	--bs-btn-color: var(--text-color-link);
+	font-style: italic;
 }
 
 .slot_button_group .btn {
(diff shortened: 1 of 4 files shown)


runbot/static/src/css/runbot.css
```diff
@@ -279,6 +286,15 @@ body, .table {
...
 }
 
+.slot_link_matched > .slot_name {
+	color: var(--text-color-link);
+	font-style: italic;
+
+}
+
+.slot_link_created > .slot_name {
+}
```
pparidans: Wut ?

runbot/static/src/css/runbot.css
```diff
@@ -279,6 +286,15 @@ body, .table {
 	padding: 0 1px;
 }
 
+.slot_link_matched > .slot_name {
+	color: var(--text-color-link);
+	font-style: italic;
+
```
pparidans: Emtpy line

runbot/static/src/css/runbot.css
```diff
@@ -169,10 +171,15 @@ dialog.modal {
...
 
-.fa {
+.fa, .fa-stack{
 	line-height: inherit; /* reset fa icon line height to body height*/
 }
 
+.fa-stack {
+    width: 1em;
+    height: 1em;
+}
```
pparidans: .fa-stack is used ? 🤔

runbot/static/src/css/runbot.css
```diff
@@ -279,6 +286,15 @@ body, .table {
 	padding: 0 1px;
 }
 
+.slot_link_matched > .slot_name {
+	color: var(--text-color-link);
```
pparidans: `--bs-btn-color` instead ?

PR #1443, opened 2026-07-03
[IMP] runbot: lazy load build options dropdown menu

pparidans:
Rendering the build options dropdown menu for every build on page load
significantly increases the initial DOM size and payload.

This commit refactors the dropdown menu to render lazily on client
interaction (hover, focus, touch, or click):
- Introduces `build-options-dropdown`, a custom HTML element that
  captures build context via `data-*` attributes.
- Adds a lightweight client-side QWeb rendering engine (`templating.js`)
  capable of handling `t-foreach`, `t-if`, `t-else`, and `{{...}}`
  interpolations using Python expression evaluation (`py_js`).
- Moves the dropdown HTML structure into client-side `<template>` tags
  (`build-options-dropdown-menu`), generating the DOM nodes only when
  the user interacts with the build menu button.


runbot/templates/utils.xml
```diff
@@ -356,83 +357,90 @@
...
-                </t>
-            </div>
+                </div>
+            </template>
+            <template id="build-options-dropdown-databases">
+                <a class="dropdown-item" href="/runbot/run/{{id}}/{{db_suffix}}">
+                    <i class="fa fa-sign-in"/>
+                    Connect {{db_suffix}}
+                </a>
+            </template>
```
pparidans: remove me

PR #1445, opened 2026-07-07
[IMP] runbot: Add a backup service to the runbot

Xavier-Do:
The template databases installed by sticky branches can be used by a lot of upgrade builds. This means that loosing one of them could break a bunch of build before it is regenerated, and even then the links between builds can persist for a while forcing to rebase the branches.

The plan to migrate in 26.04 will erase all servers in fine, making this problem more visible.

This pr introduces a new backup service that allows to give a fallback for a given database. All templates of sticky branches will be stored there, and the restore step will fallback on this server if the database is not found on the initial host.

### DIFF
--- a/runbot/models/runbot.py
+++ b/runbot/models/runbot.py
@@ -7,7 +7,7 @@ import signal
 import subprocess
 import time
 from contextlib import contextmanager
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import docker
 from requests.exceptions import HTTPError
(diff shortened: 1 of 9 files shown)


runbot/models/runbot.py
```diff
@@ -365,6 +365,33 @@ def _start_docker_registry(self):
...
+            assert str(database.build_id.id) in database.name  # ensure that database are well uniquified by build in order to avoid to forget to adapt this if it changes in the future
+            try:
+                subprocess.run(['wget', '-O', os.path.join(backup_location, f'{database.name}.zip'), url], check=True)
+            except subprocess.CalledProcessError as e:
+                _logger.error('Failed to backup database %s: %s', database.name, e)
+        # cleanup old backup
+        # we could keep them longer than on hosts but let's keep it simple for now
+        for remaining_backup in existing_backups:
+            _logger.info('Removing old backup %s', remaining_backup)
+            os.remove(self._root('backups', f'{remaining_backup}.zip'))
```
d-fence:
```suggestion
            os.remove(self._path('backups', f'{remaining_backup}.zip'))
```

d-fence (changes requested):
otherwise LGTM

PR #1446, opened 2026-07-07
[IMP] runbot: allow to disable hosts selectively

d-fence:


Xavier-Do:
I'm not sure that the action to pause all host will work because of concurrency but we could solve that another way
The global pause_all_hosts should work but having it in the settings views would be great

PR #1455, opened 2026-07-16
[IMP] runbot: use proper version names in HistoryGraph

hubvd:
Instead of displaying "~" or "19.04", we now show "master" or "saas-19.4"


Xavier-Do:
I'm not a big fan of the idea, version number can be ordered, and it has a fixed length, why do you want to change that?

hubvd:
image

This is only for the used in the graph tooltip no ? 
When showing the error page to newcomers, it is not exactly clear for them that "~" means master in this case

Xavier-Do:
The fixed length for stable vs saas version is still great IMHO because when you move your mouse around, the tool-tip doesn't move and resize too much. 
But I understand that this technical naming can be confusing so lets test this version.

PR #1457, opened 2026-07-20
[FIX] git-fw: clear stale commit map on redo

hadielyakhni:
`redo` writes `.git/FW_CMAP` but never removes it, so the next `redo` may read the map left by a previous one and not find its commit in it.

Remove the file at the start of `redo`.


xmo-odoo:
Independently implemented after report from a colleague.

PR #1477, opened 2026-08-03
Reference slot instead of build

Xavier-Do:
This would allow to create upgrade build before the template is finished

The main issue being that it removes all the logic currently used for upgrades
```
  dump_builds = build.search([('id', 'child_of', source.id), ('params_id.config_id', '=', config_id.id), ('orphan_result', '=', False)])
                    # this search is not optimal
                    if not dump_builds:
                        build._log('_run_configure_upgrade', 'No build found with config %s in %s' % (config_id.name, source.id), level='ERROR')
                    dbs = dump_builds.database_ids.sorted('db_suffix')
                    valid_databases += list(self._filter_upgrade_database(dbs, upgrade_db.db_pattern))
                    if not valid_databases:
                        build._log('_run_configure_upgrade', 'No database found for pattern %s' % (upgrade_db.db_pattern), level='ERROR')

                for db in valid_databases:
                    child = build._add_child({
                        'upgrade_to_build_id': None,
```
    
Not correct in current state

This solution has a main drawback: the build id is the right way to uniquify the parameters of the upgrade since it defines where the database will be downloaded. 
Using the slot id is correct but will potentially break some linking since two slot can reference the same build. 
Using the parameter is not 100% the same since we may rebuild the build and have a new database for the same params. So the linking could not be done.

### DIFF
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -106,6 +106,7 @@ class BuildParameters(models.Model):
 
     upgrade_to_build_id = fields.Many2one('runbot.build', index=True)  # use to define sources to use with upgrade script
     upgrade_from_build_id = fields.Many2one('runbot.build', index=True)  # use to download db
+    upgrade_from_slot_id = fields.Many2one('runbot.batch.slot', index=True)  # replaces upgrade_from_build_id when build is not created yet
     dump_db = fields.Many2one('runbot.database', index=True)  # use to define db to download
 
     fingerprint = fields.Char('Fingerprint', compute='_compute_fingerprint', store=True, index=True)
@@ -1395,7 +1396,7 @@ class BuildResult(models.Model):
                 if os.path.isdir(commit._source_path(upgrade_path)):
                     yield os.sep.join([repo_folder, upgrade_path]).strip(os.sep)
 
-    def _upgrade_builds_references(self, refs_batches=None):
+    def _upgrade_slot_references(self, refs_batches=None):
         params = self.params_id
         params.ensure_one()
         trigger = params.trigger_id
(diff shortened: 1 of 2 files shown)


Xavier-Do:
This may need to make link child optionnal
```patch
diff --git a/runbot/models/build.py b/runbot/models/build.py
index ecb76275..a1f90ce8 100644
--- a/runbot/models/build.py
+++ b/runbot/models/build.py
@@ -410,11 +410,14 @@ class BuildResult(models.Model):
             children_ids += [link.child_id.id for link in record.child_link_ids if not link.orphan_result and not link.child_id.orphan_result]
             children = self.browse(children_ids)
             if record._get_state_score(record.local_state) > waiting_score and children:  # if finish, check children
-                children_state = record._get_youngest_state([child.global_state for child in children])
-                if record._get_state_score(children_state) > waiting_score:
-                    record.global_state = record.local_state
-                else:
+                if any(not link.child_id for link in record.child_link_ids):
                     record.global_state = 'waiting'
+                else:
+                    children_state = record._get_youngest_state([child.global_state for child in children])
+                    if record._get_state_score(children_state) > waiting_score:
+                        record.global_state = record.local_state
+                    else:
+                        record.global_state = 'waiting'
             else:
                 record.global_state = record.local_state
 
@@ -1878,15 +1881,18 @@ class BuildLink(models.Model):
     _order = 'id desc'
 
     parent_id = fields.Many2one('runbot.build', string='Parent Build', required=True, ondelete='cascade')
-    child_id = fields.Many2one('runbot.build', string='Child Build', required=True, ondelete='cascade')
-    params_id = fields.Many2one('runbot.build.params', string='Params', related='child_id.params_id', store=True)
+    child_id = fields.Many2one('runbot.build', string='Child Build', ondelete='cascade')
+    params_id = fields.Many2one('runbot.build.params', string='Params', required=True)
     orphan_result = fields.Boolean(string='Orphan Result', help='If set, the result of the child build will not be taken into account for the parent build result')
 
     _no_self_link = models.Constraint('check (parent_id != child_id)', "a build cannot link itself")
-    _unique_link = models.Constraint('unique (parent_id, child_id)', "duplicate build link")
+    _unique_link = models.Constraint('unique (parent_id, params_id, child_id)', "duplicate build link")
 
     @api.model_create_multi
     def create(self, vals_list):
+        for vals in vals_list:
+            if vals.get('child_id'):
+                vals['params_id'] = self.env['runbot.build'].browse(vals.get('child_id')).params_id.id
         links = super().create(vals_list)
         links.mapped('parent_id')._update_globals()
         return links

```

PR #1484, opened 2026-08-11
EXAMPLE AI PR, FEEL FREE TO DELETE, RE-USE ....

jeanschoenlaub:
[IMP] runbot: restore a dump from any build of the current batch

When a custom trigger uses `restore_mode = auto`, the dump is looked up through `dump_trigger` in the base reference batch. That batch is only set when the merge base of the bundle is the head of a batch of the base bundle, which is not the case when starting a build from an arbitrary commit, e.g. to bisect an undeterministic error.

The lookup then returns an empty recordset, the `if reference_build:` block is silently skipped, and the restore falls back on `reference_build_id`. When that build only creates children it never created a database, so the composed url points to a dump that never existed: the restore fails on a 404 and the following steps report an unhelpful `Database not initialized`.

Fall back on any finished build of the current batch holding a database with the expected suffix. A dump producing trigger started manually to prepare the restore is not necessarily `dump_trigger`, but the database it builds is just as good. Only builds in `done` or `running` are considered, the dump of a build still testing would be incomplete.

### DIFF
--- a/runbot/models/build_config.py
+++ b/runbot/models/build_config.py
@@ -1168,6 +1168,22 @@ class ConfigStep(models.Model):
                     build._log('_run_restore', f'No dump with suffix {dump_suffix} found in build [{reference_build.id}]({reference_build.build_url})', log_type='markdown', level='ERROR')
                     build._kill(result='ko')
                     return
+            if not dump_db:
+                # No reference build was found, e.g. because the bundle is
+                # based on a commit that is not the head of a base batch and
+                # has therefore no base reference batch. A dump producing
+                # trigger started manually to prepare the restore is not
+                # necessarily `dump_trigger`, but it sits in the current batch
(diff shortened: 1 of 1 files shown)


Xavier-Do:
Regarding the pr title it almost looks like this pr should be ignored so i won't spend a lot of time on it ;)

It looks like it adds a lot of complexity and may breaks some flow
The logic doesn't really make sense  to me. Taking a random finished database looks like the best idea to completely loose track on whats going on and have unpreditctable behaviour. 
It could also break the fallback on reference_build_id 

You can come and talk to us about your needs if you want

jeanschoenlaub:
Thanks for taking the time, the need is to be able to use the runbot to do a bisect with specific commits, see:
https://discord.com/channels/678381219515465750/704997651833159831/1536640736236146728

(100% if you don't implement, just suggestion)

PR #1501, opened 2026-08-21
[IMP] runbot: store diff on CommitLink

d-fence:
When fetching commit info, store the diff as well so that it can be reused later in some build config steps and save time. A mechanism should be added to clean up old diffs.

### DIFF
--- a/runbot/models/batch.py
+++ b/runbot/models/batch.py
@@ -529,9 +529,9 @@ class Batch(models.Model):
                     continue
 
                 # diff. Iter on --numstat, easier to parse than --shortstat summary
-                diff = commit.repo_id._git(['diff', '--numstat', merge_base_sha, commit.name]).strip()
-                if diff:
-                    for line in diff.split('\n'):
+                diff_stat = commit.repo_id._git(['diff', '--numstat', merge_base_sha, commit.name]).strip()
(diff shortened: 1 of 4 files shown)


runbot/models/batch.py
```diff
@@ -526,16 +526,17 @@ def _update_commits_infos(self, base_head_per_repo):
...
+                if diff_stat:
+                    for line in diff_stat.split('\n'):
                         link_commit.file_changed += 1
                         add, remove, _ = line.split(None, 2)
                         try:
                             link_commit.diff_add += int(add)
                             link_commit.diff_remove += int(remove)
                         except ValueError:  # binary files
                             pass
+                link_commit.diff = commit.repo_id._git(['diff', f'{merge_base_sha}..{commit.name}', '--', '*'])
```
Xavier-Do:
```suggestion
                if link_commit.file_changed <= self.bundle_id.file_limit and link_commit.base_ahead <= bundle.commit_limit:
                    link_commit.diff = commit.repo_id._git(['diff', f'{merge_base_sha}..{commit.name}', '--', '*'])
```

d-fence: Good idea, we avoid storing gigantic diff's when a PR is wrongly configured

PR #1522, opened 2026-09-02
[FIX] runbot: revert 4c245ac

Syozik:
Commit [1] made the history graph blurry and broke the hovering of the
cells.
Instead of that fix, we move the widget to the next row, this way it
shouldn't be overflown.

[1]: https://github.com/odoo/runbot/commit/4c245ac9e8c2f251c121db7c6c7e18b9662ab8a0

### DIFF
--- a/runbot/static/src/js/fields/history_graph.js
+++ b/runbot/static/src/js/fields/history_graph.js
@@ -3,8 +3,8 @@ import { useRef, xml, Component, useEffect } from "@odoo/owl";
 
 export class HistoryGraph extends Component {
     static template = xml`
-        <div>
-            <canvas class="w-100" t-ref="canvas"/>
+        <div class="w-100 overflow-auto">
+            <canvas t-ref="canvas"/>
         </div>
     `;
 
--- a/runbot/views/build_error_views.xml
+++ b/runbot/views/build_error_views.xml
@@ -63,7 +63,8 @@
                   <field name="last_seen_date" widget="frontend_url" options="{'link_field': 'last_seen_build_id', 'numeric': True}"/>
                   <field name="first_seen_build_id" invisible="True"/>
                   <field name="last_seen_build_id" invisible="True"/>
-                  <field name="history_data" widget="history_graph" options="{'cell_size': 13, 'mouse_actions': True}"/>
+                  <label for="history_data" colspan="2"/>
+                  <field name="history_data" nolabel="1" colspan="2" widget="history_graph" options="{'cell_size': 13, 'mouse_actions': True}"/>
                   <button name="action_appearing_batches" string="Appearing Batches" type="object" class="btn btn-primary" invisible="not first_seen_batch_ids"/>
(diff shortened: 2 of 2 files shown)


Syozik:
Hi @Xavier-Do :wave: I accidentally broke the history graph's cell hovering (works fine without being zoomed in, but can be seen when zoomed in, or with a smaller screen size)... Could you take a look at this PR?
Sorry about it...

Xavier-Do:
Yes we noticed, it was not a big deal but I was thinking to ask you to fix it, so, thanks for the pr. I'll let @pparidans review :)

pparidans (changes requested):
Hi,

This commit introduces a regression in the errors' list view:

Image

I don't think that the initial fix in https://github.com/odoo/runbot/pull/1519 was the right one : by moving the `w-100` to the canvas, it renders at a lower scale (depending on its parent's size) then stretches it to fit the 100% width.

The current commit makes it render at an even lower scale.

Xavier-Do:
I wonder if we should just move the widget in a single column in the form view to have more place and even allow to increase the size

Syozik:
Hey @pparidans, can I just revert my previous commit and simply render the history graph with a scrollbar? Not the best, but better than it was before? 
image

pparidans:
> Hey @pparidans, can I just revert my previous commit and simply render the history graph with a scrollbar? Not the best, but better than it was before? 

I tried this solution as well... but I'm not that fond of not seeing the more recent (right side) columns of the graph without scrolling...

Syozik:
I thought about 2 solutions to have it auto-scrolled to the right side:
- using display flex and direction row-reverse (hacky, but it will be auto-scrolled)
- doing  `this.canvasRef.el.scrollLeft = this.canvasRef.el.scrollWidth;` after the first render
Does the 2nd solution sound like an appropriate one?

Syozik:
@pparidans  Here's the first solution:  

https://github.com/user-attachments/assets/a182417a-b841-401c-b72a-1d8e20e3527d

pparidans:
> I thought about 2 solutions to have it auto-scrolled to the right side:
> 
> * using display flex and direction row-reverse (hacky, but it will be auto-scrolled)
> * doing  `this.canvasRef.el.scrollLeft = this.canvasRef.el.scrollWidth;` after the first render
>   Does the 2nd solution sound like an appropriate one?

Honestly, both feels "hackish" to me and I would avoid them if possible...

As suggested by @Xavier-Do , I think that giving more space to the graph by moving its label above would solve most of the issue (eventually keeping the overflow just in case).

pparidans:
Any reason you didn't kept the overflow (as a fallback) ?

Also, could you write a proper commit message (title!)

Syozik:
@pparidans oh, it's my bad, I misinterpreted `eventually keeping the overflow`,
I added `overflow: auto` back and changed the commit message title. Thanks for checking it :bow: 
Good to go?

PR #1532, opened 2026-09-14
[IMP] runbot: merge legacy theme into light

pparidans:
The legacy theme duplicated the light theme in all but a couple of custom rules, and every runbot install was still cookied to legacy by default. Fold the legacy-specific overrides into the light theme, drop the now-redundant Legacy option from the theme picker, and migrate any stale theme=legacy cookies to light so existing users transition without a visual change.

While at it, trim the light theme's .btn-success override down to the two properties that actually differ from Bootstrap's defaults, and extend the text-bg-info contrast fix to the red404 theme, which shares the same custom info color and had the same readability issue.


pparidans:
**ci/security**: `qcontext['theme']` was already present ; only handling the legacy/light change.

PR #1537, opened 2026-09-16
[IMP] runbot: only compute build error counters when needed

pparidans:
The navbar build error link is only rendered for logged in users but the three search_count were executed for every frontend request, including anonymous ones which are the bulk of the traffic. Move them to a helper called while filling the qcontext, skip them for public users, and only count all the errors when it is actually used as a fallback.


pparidans:
Not needed if https://github.com/odoo/runbot/pull/1539 is merged.

